PR #216, opened 2015-01-29
Add build instructions to readme and make InjectiveSerializer serializable

rmetzger:
Hi,

Apache Flink recently started using chill and Kryo as a fallback serializer. We really like the work done here.
Since we are sending the serializers around in the cluster, they need to be Serializable. I found that the `InjectiveSerializer` is not Serializable, so I made it implement the interface.

I also added some build instructions to the readme, because it took me some time to research the sbt commands.

Please let me know if you want something in this pull request to be solved differently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,16 @@ Extensions for the [Kryo serialization library](https://github.com/EsotericSoftw
 serializers and a set of classes to ease configuration of Kryo in systems like Hadoop, Storm,
 Akka, etc.
 
+
+### Buidling Chill
+
+```bash
+./sbt
+> compile # to build chill
+> publishM2 # to publish chill to your local .m2 repo
+> publish-local # publish to local ivy repo.
+```
+
 Chill has a set of subprojects: chill-java, chill-hadoop, chill-storm and chill-scala.  Other than
 chill-scala, all these projects are written in Java so they are easy to use on any JVM platform.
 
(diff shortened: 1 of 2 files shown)


ianoc:
Generally a lot of serializers may not be themselves serializable, the approach chill usually takes here is that you serialize with kryo the ability to build the serializer. https://github.com/twitter/chill/blob/develop/chill-hadoop/src/main/java/com/twitter/chill/hadoop/KryoSerialization.java


Though i've no real problem with making this class serializable. Little odd the sbt stuff given its the most common scala build tool. But don't really mind either

johnynek:
Second what Ian said about us generally avoiding serializing serializers. Since Kryo is mutable, we generally use `KryoInstantiator` instances to create new ones.  Also there is `KryoPool` so you can safely share instances without allocating each time you need one.

ConfiguredInstantiator is one way we solve this problem across a lot of use cases.

rmetzger:
Thank you for merging the pull request so quickly.

We usually don't need Kryo serializers to be Serializable because we're sending them around as Classes which Kryo will instantiate. 
But this particular serializer needed to be serializable because its used by the [AvroSerializer](https://github.com/twitter/chill/blob/develop/chill-avro/src/main/scala/com/twitter/chill/avro/AvroSerializer.scala) and there you need to pass the avro schema to it.

I'll take a closer look into the `ConfiguredInstantiator`.

Issue #218, opened 2015-02-13
there's no chill-scrooge for scala 2.11


colinmarc:
It's disabled in Build.scala - why?

ianoc:
At the time we did the update/release there was no scrooge serializer package for 2.11, would welcome a PR if thats changed since then

PR #219, opened 2015-02-13
Build chill-scrooge for scala 2.11, too

colinmarc:
I'm not sure why it was disabled - I was able to build with no issues.


colinmarc:
(should fix #218)

ianoc:
It was disabled because it wasn't published/the jar didn't exist :)

ianoc:
Thanks for the fix

PR #220, opened 2015-02-23
Rewrite Java BitSet serializer to make it more efficient

rmetzger:
The new serializer is not writing an entire byte per bit.
To access the internal state of BitSet, I have to use Java reflection. The Java 1.7 implementation of BitSet allows to access the internal long[], but chill is compatible to Java 1.6, so we have to play some tricks here.

My experiments show that the new code is 4 times more space efficient (514 vs 2052 bytes) and 11 times faster (446 vs 4908 ms) [both in 100000 ser-deser rounds]

### DIFF
--- a/chill-java/src/main/java/com/twitter/chill/java/BitSetSerializer.java
+++ b/chill-java/src/main/java/com/twitter/chill/java/BitSetSerializer.java
@@ -17,6 +17,7 @@ limitations under the License.
 package com.twitter.chill.java;
 
 import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.KryoException;
 import com.esotericsoftware.kryo.Serializer;
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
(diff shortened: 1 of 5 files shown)


chill-java/src/main/java/com/twitter/chill/java/BitSetSerializer.java
```diff
@@ -24,32 +25,80 @@
...
+            words = (long[]) wordsField.get(bitSet);
+        } catch (IllegalAccessException e) {
+            throw new KryoException("Error while accessing field 'words' of bitSet", e);
+        }
+        output.writeInt(words.length, true);
 
-        for(int i = 0; i < len; i++) {
-            output.writeBoolean(bitSet.get(i));
+        for(int i = 0; i < words.length; i++) {
+            output.writeLong(words[i]);
```
johnynek: are these numbers positive? If so, `writeLong(words[i], true)` might be a win.

rmetzger: No, they are not. The `words` contain the the bits from the bitset. Depending on how the user sets the bits, they might represent very negative numbers ;)

chill-java/src/test/scala/com/twitter/chill/java/BitSetTest.scala
```diff
@@ -0,0 +1,91 @@
...
+      println("The new serializer took "+(System.currentTimeMillis() - start)+"ms")
+
+      var out = new Output(1, -1)
+      oldKryo.writeObject(out, element)
+      out.flush()
+      println("The old serializer needs "+out.total()+" bytes")
+      out = new Output(1, -1)
+      newKryo.writeObject(out, element)
+      out.flush()
+      println("The new serializer needs "+out.total()+" bytes")
```
johnynek: can you write this as a test? The new bytes must be `<=` the old bytes?

chill-java/src/test/scala/com/twitter/chill/java/BitSetTest.scala
```diff
@@ -0,0 +1,91 @@
...
+
+      val element = new util.BitSet(2048)
+      val rnd = new Random()
+      for (i <- 0 to 2048) {
+        element.set(i, rnd.nextBoolean())
+      }
+
+      // warmup In case anybody wants to see hotspot
+      for(i <- 0 to 50000) {
+        rt(element)(oldKryo)
```
johnynek: can you add a test that the random bitset is the same in the old and new cases?

rmetzger:
Thank you for reviewing the pull request.
I'll address your comments tomorrow.

chill-java/src/main/java/com/twitter/chill/java/BitSetSerializer.java
johnynek: Okay. I didn't know if they stored deltas to the next present value.

rmetzger:
I've updated the pull request.
The test now checks the contents of the random BitSet against the new and old serializer contents.

johnynek:
Thank you, sir!

rmetzger:
Thank you for merging the change ;)

PR #222, opened 2015-03-18
Explicitly add Stream support (which seemed to work before)

johnynek:
even without registering, it looks like the FieldsSerializer will work in many common cases. Also, you can register on your own using the same line we added here.


ianoc:
LGTM, merge when green

johnynek:
I'm actually skeptical about this now. It seems that serializing streams ofter works already, but serializes the closure. Some people may be relying on this and changing the default seems like it might be a bad behavior, especially considering how easy it is to add support for with a one-line change to your KryoInstantiator.

PR #229, opened 2015-06-05
Alter scala collection default serializer registration order

nonvirtualthunk:
In the event that a class is not registered and Kryo falls back on implicit registration it will check the registered default serializers in the order they were registered until it finds one that could apply. Currently in ScalaKryoInstantiator the more general Buffer appears before ListBuffer, so when serializing/deserializing a ListBuffer in this situation, the Buffer serializer will be chosen, and you will end up with an ArrayBuffer on the receiving end, potentially occupying a variable the JVM is convinced is a ListBuffer, which leads to serious problems. This change simply ensures that ListBuffer is registered first.


ianoc:
Thanks for the fix and detailed explanation of the issue.

PR #234, opened 2015-07-12
Version Upgrade & Cleanup

reid-spencer:
- Chill: 0.6.0 -> 0.7.0-SNAPSHOT
- SBT: 0.13.5 -> 0.13.8
- Kryo: 2.21 -> 3.0.2
- Scala: 2.11.5 -> 2.11.7
- Akka-Core: 2.3.6 -> 2.3.11
- Bijection-Core: 0.8.0 -> 0.8.1
- Typesafe Config: 1.2.1 -> 1.3.0
- Java: Remove restriction on 1.6 target bytecode (might break Hadoop?)
- Change the organization name for the kryo package for the change in the 3.0 release.
- Use the kryo-shaded library
- Add -Xlint:unchecked to identify numerous unchecked warnings
- Avoid some long->int truncations
- Deal with ObjectInstantiator now having type arguments
- Clean up JavaDoc so it doesn't fail

FYI: For me this passes all tests and publishes locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: scala
 sudo: false
 scala:
-  - 2.10.4
-  - 2.11.2
+  - 2.11.7
+jdk:
+   - oraclejdk8
+
--- a/chill-java/src/test/scala/com/twitter/chill/java/BitSetTest.scala
+++ b/chill-java/src/test/scala/com/twitter/chill/java/BitSetTest.scala
@@ -3,7 +3,7 @@ package com.twitter.chill.java
 import java.util
 
 import com.esotericsoftware.kryo.Kryo
-import com.esotericsoftware.kryo.io.{Input, Output}
+import com.esotericsoftware.kryo.io.{ Input, Output }
 import org.objenesis.strategy.StdInstantiatorStrategy
 import org.scalatest._
 
@@ -63,19 +63,19 @@ class BitSetSpec extends WordSpec with MustMatchers {
       }
 
       // warmup In case anybody wants to see hotspot
-      var lastBitSetFromOld : util.BitSet = null
-      for(i <- 0 to 50000) {
+      var lastBitSetFromOld: util.BitSet = null
+      for (i <- 0 to 50000) {
         lastBitSetFromOld = rt(element)(oldKryo)
       }
       var start = System.currentTimeMillis()
-      for(i <- 0 to 100000) {
+      for (i <- 0 to 100000) {
         rt(element)(oldKryo)
       }
-      println("The old serializer took "+(System.currentTimeMillis() - start)+"ms")
+      println("The old serializer took " + (System.currentTimeMillis() - start) + "ms")
 
-      var lastBitSetFromNew : util.BitSet = null
+      var lastBitSetFromNew: util.BitSet = null
       // warmup for the new kryo
-      for(i <- 0 to 50000) {
+      for (i <- 0 to 50000) {
         lastBitSetFromNew = rt(element)(newKryo)
       }
       // check for the three bitsets to be equal
@@ -87,23 +87,22 @@ class BitSetSpec extends WordSpec with MustMatchers {
         element.get(i) must be(lastBitSetFromNew.get(i))
       }
 
-
       start = System.currentTimeMillis()
-      for(i <- 0 to 100000) {
+      for (i <- 0 to 100000) {
         rt(element)(newKryo)
       }
-      println("The new serializer took "+(System.currentTimeMillis() - start)+"ms")
+      println("The new serializer took " + (System.currentTimeMillis() - start) + "ms")
 
       var out = new Output(1, -1)
       oldKryo.writeObject(out, element)
       out.flush()
       var oldBytes = out.total()
-      println("The old serializer needs "+oldBytes+" bytes")
+      println("The old serializer needs " + oldBytes + " bytes")
       out = new Output(1, -1)
       newKryo.writeObject(out, element)
       out.flush()
       var newBytes = out.total()
-      println("The new serializer needs "+newBytes+" bytes")
+      println("The new serializer needs " + newBytes + " bytes")
 
       oldBytes >= newBytes must be(true)
     }
--- a/chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
+++ b/chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
@@ -76,10 +76,10 @@ class KryoBase extends Kryo {
   def tryStrategy(cls: Class[_]): InstantiatorStrategy =
     strategy.getOrElse {
       val name = cls.getName
-      if (cls.isMemberClass() && !Modifier.isStatic(cls.getModifiers()))
-        throw new KryoException("Class cannot be created (non-static member class): " + name);
+      if (cls.isMemberClass && !Modifier.isStatic(cls.getModifiers))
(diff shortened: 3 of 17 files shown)


ianoc:
Big question for these is largely why for a lot of them. Why loose 1.6
compatibility ? Why change Kryp version ? These are going to make this
update hard to consume downstream and cause a lot of problems for people.
So be good to have strong motivation for anything that will break binary
compatibility or serialization comparability.

reid-spencer:
@ianoc - The main reason is upgrading to Kryo 3.0 because I needed to use its Java 8 features and bug fixes.  In my world Java 6 and 7 are dead and so is Scala 2.10.  I also wanted to be sure the akka version was the same as what I'm using. So, since I was updating things, I decided to go all out. This patch PR doesn't need to be useful to you, but I provided it in case you do want to go in this direction.  I'm fine working on my fork until such time as Chill modernizes itself.  By way of compromise, which of the changes  do you find most objectionable? I might be willing to back some of these (not necessarily required changes) out. I do require Kryo 3.0.2, sbt 0.13.8, scala 2.11.7 and JVM 8 support, however.

FYI: The Java 6 support is only there for Hadoop (or so the comment says) and the Hadoop version used here is WOEFULLY out of date :) Maybe its time to upgrade anyway?

.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: scala
 sudo: false
 scala:
-  - 2.10.4
```
johnynek: we shouldn't drop scala 2.10 support. Twitter (and other big companies take a long time to upgrade such things). We need our tests to run on the version we use.

chill-java/src/test/scala/com/twitter/chill/java/BitSetTest.scala
```diff
@@ -87,23 +87,22 @@ class BitSetSpec extends WordSpec with MustMatchers {
...
       }
-      println("The new serializer took "+(System.currentTimeMillis() - start)+"ms")
+      println("The new serializer took " + (System.currentTimeMillis() - start) + "ms")
 
       var out = new Output(1, -1)
       oldKryo.writeObject(out, element)
       out.flush()
       var oldBytes = out.total()
-      println("The old serializer needs "+oldBytes+" bytes")
+      println("The old serializer needs " + oldBytes + " bytes")
```
johnynek: while we are cleaning up, can we use string interpolation?

chill-java/src/test/scala/com/twitter/chill/java/BitSetTest.scala
```diff
@@ -87,23 +87,22 @@ class BitSetSpec extends WordSpec with MustMatchers {
...
       out = new Output(1, -1)
       newKryo.writeObject(out, element)
       out.flush()
       var newBytes = out.total()
-      println("The new serializer needs "+newBytes+" bytes")
+      println("The new serializer needs " + newBytes + " bytes")
```
johnynek: interpolate

chill-java/src/test/scala/com/twitter/chill/java/BitSetTest.scala
```diff
@@ -87,23 +87,22 @@ class BitSetSpec extends WordSpec with MustMatchers {
...
       }
 
-
       start = System.currentTimeMillis()
-      for(i <- 0 to 100000) {
+      for (i <- 0 to 100000) {
         rt(element)(newKryo)
       }
-      println("The new serializer took "+(System.currentTimeMillis() - start)+"ms")
+      println("The new serializer took " + (System.currentTimeMillis() - start) + "ms")
```
johnynek: while we are cleaning up, can we use string interpolation?

chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
```diff
@@ -102,8 +102,8 @@ class KryoBase extends Kryo {
 object Instantiators {
   // Go through the list and use the first that works
   def newOrElse(cls: Class[_],
```
johnynek: let's add a type here if we can. Can't these all be the same type?
```scala
def newOrElse[K](cls: Class[_ <: K],
  it: TraversableOnce[Class[_ <: K] => Try[ObjectInstatiator[K]],
  else: => ObjectInstatiator[K]): ObjectInstatiator[K] = {
```
should be doable, right? It would be better to have some types here.

chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
```diff
@@ -112,15 +112,15 @@ object Instantiators {
       }
     }
```
johnynek: `.collectFirst { case Some(x) => x }` is better than `find.flatten`

chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
```diff
@@ -130,7 +130,7 @@ object Instantiators {
...
 
   // This one tries reflectasm, which is a fast way of constructing an object
-  def reflectAsm(t: Class[_]): Either[Throwable, ObjectInstantiator] = {
+  def reflectAsm(t: Class[_]): Either[Throwable, ObjectInstantiator[_]] = {
```
johnynek: same `t: Class[_ <: K]` should do it, and we can convert to `Try[ObjectInstatiator[K]]` here. There was no `Try` when we wrote this.

chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
```diff
@@ -154,7 +154,7 @@ object Instantiators {
     }
   }
 
-  def normalJava(t: Class[_]): Either[Throwable, ObjectInstantiator] = {
+  def normalJava(t: Class[_]): Either[Throwable, ObjectInstantiator[_]] = {
```
johnynek: add the type, switch to Try.

project/Build.scala
```diff
@@ -18,27 +16,25 @@ object ChillBuild extends Build {
...
       Opts.resolver.sonatypeSnapshots,
       Opts.resolver.sonatypeReleases
     ),
     libraryDependencies ++= Seq(
-      "org.scalacheck" %% "scalacheck" % "1.11.5" % "test",
-      "org.scalatest" %% "scalatest" % "2.2.2" % "test",
-      "com.esotericsoftware.kryo" % "kryo" % kryoVersion
+      "org.scalacheck" %% "scalacheck" % "1.12.4" % "test",
+      "org.scalatest" %% "scalatest" % "2.2.5" % "test",
+      "com.esotericsoftware" % "kryo-shaded" % kryoVersion
```
johnynek: what is the shaded jar? Can you comment on what is shaded here?

project/Build.scala
```diff
@@ -185,8 +181,8 @@ object ChillBuild extends Build {
     autoScalaLibrary := false,
     libraryDependencies ++= Seq(
       "org.apache.hadoop" % "hadoop-core" % "0.20.2" % "provided",
-      "org.slf4j" % "slf4j-api" % "1.6.6",
-      "org.slf4j" % "slf4j-log4j12" % "1.6.6" % "provided"
+      "org.slf4j" % "slf4j-api" % "1.7.12",
```
johnynek: why are you bumping these up? Do we need it? This going to force it on everyone downstream (which includes scalding, spark, summingbird, and a lot of people internal at Twitter).

ianoc:
no updates in a long time, we've upgraded to latest kryo. Keeping 2.10 and older java support for now. Closing

PR #235, opened 2015-07-25
SketchMap Serializer

alekseyig:
Hello, 
In our system we store the SketchMap as a blob in Postgres, to do so we need somehow serialize it into binary, initially we used the standard Kryo serializer - FieldSerializer. It worked, but after we tried to upgrade our dependence on algebird from 0.4 to 0.6 the deserialization failed. The internal implementation of SketchMap changed and  FieldSerializer was not able to map stored binaries back to new SketchMap. We needed to come up with a serialization that would survive betweens the upgrades. 

Proposed is SketchMap Serializer that depends on public Algebird API, generic to types of key and values and has a small footprint (it tries to serialize only the non zero value bits).

Thank you for your consideration. 

P.S. The proposed serializer is currently used in our production system.

### DIFF
--- a/chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
+++ b/chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
@@ -94,14 +94,14 @@ class SketchMapSerializer[K, V](implicit skmMonoid: SketchMapMonoid[K, V], value
     val rows = skm.valuesTable.rows
     val cols = skm.valuesTable.cols
 
-    val values: IndexedSeq[(Int, Int, V)] =
+    val values: Seq[((Int, Int), V)] =
       for (
-        r: Int <- 0 to rows - 1; c: Int <- 0 to cols - 1 if skm.valuesTable.getValue((r, c)) != valueMonoid.zero
-      ) yield (r, c, skm.valuesTable.getValue((r, c)))
+        r: Int <- 0 until rows; c: Int <- 0 until cols if valueMonoid.isNonZero(skm.valuesTable.getValue((r, c)))
+      ) yield ((r, c), skm.valuesTable.getValue((r, c)))
 
     output.writeInt(rows, true)
     output.writeInt(cols, true)
-    kryo.writeClassAndObject(output, values.toList)
+    kryo.writeClassAndObject(output, values)
     kryo.writeClassAndObject(output, skm.totalValue)
     kryo.writeClassAndObject(output, skm.heavyHitterKeys)
 
@@ -111,7 +111,7 @@ class SketchMapSerializer[K, V](implicit skmMonoid: SketchMapMonoid[K, V], value
   def read(kryo: Kryo, in: Input, cls: Class[SketchMap[K, V]]): SketchMap[K, V] = {
     val rowsOrig = in.readInt(true)
     val colsOrig = in.readInt(true)
-    val values = kryo.readClassAndObject(in).asInstanceOf[List[(Int, Int, V)]]
+    val values = kryo.readClassAndObject(in).asInstanceOf[Seq[((Int, Int), V)]]
     val totalValue = kryo.readClassAndObject(in).asInstanceOf[V]
     val heavyHitterKeys = kryo.readClassAndObject(in).asInstanceOf[List[K]]
 
@@ -119,7 +119,7 @@ class SketchMapSerializer[K, V](implicit skmMonoid: SketchMapMonoid[K, V], value
     val cols = if (colsOrig == 0) skmMonoid.params.width else colsOrig
 
     val zero = AdaptiveMatrix.fill[V](rows, cols)(valueMonoid.zero)
-    val valuesTable = values.foldLeft(zero){ case (acc, (r, c, v)) ⇒ acc.updated((r, c), v) }
+    val valuesTable = values.foldLeft(zero){ case (acc, ((r, c), v)) => acc.updated((r, c), v) }
 
     SketchMap(valuesTable, heavyHitterKeys, totalValue)
   }
--- a/chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
+++ b/chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
@@ -19,20 +19,11 @@ import com.esotericsoftware.kryo.Kryo
 import com.esotericsoftware.kryo.{ Serializer => KSerializer }
 import com.esotericsoftware.kryo.io.{ Input, Output }
 
-import com.twitter.algebird.{
-  AveragedValue,
-  DecayedValue,
-  HLL,
-  HyperLogLog,
-  HyperLogLogMonoid,
-  Moments,
-  SpaceSaver,
-  SSOne,
-  SSMany
-}
+import com.twitter.algebird._
+import com.twitter.algebird.matrix.AdaptiveMatrix
 
 import scala.collection.mutable.{ Map => MMap }
-import scala.collection.immutable.SortedMap
+
 
 class AveragedValueSerializer extends KSerializer[AveragedValue] {
   setImmutable(true)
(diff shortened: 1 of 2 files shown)


chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
```diff
@@ -95,3 +86,42 @@ class HLLMonoidSerializer extends KSerializer[HyperLogLogMonoid] {
...
+    val cols = skm.valuesTable.cols
+
+    val values: IndexedSeq[(Int, Int, V)] =
+      for (
+        r: Int <- 0 to rows - 1; c: Int <- 0 to cols - 1 if skm.valuesTable.getValue((r, c)) != valueMonoid.zero
+      ) yield (r, c, skm.valuesTable.getValue((r, c)))
+
+    output.writeInt(rows, true)
+    output.writeInt(cols, true)
+    kryo.writeClassAndObject(output, values.toList)
```
johnynek: is the toList needed there? That is going to be a full iteration through things again.

A few notes:
1. you could use `0 until rows`. rather than the `-1`.
2. map/flatMap on Range is not lazy, so this is going to fully materialize the filtered table twice, once in the for expression, once in .toList.
3. prefer valueMonoid.isNonZero rather than the equality check (it's janky in the end, and I wish we had a good Equivalence typeclass, but its what exists now).

johnynek: 1. Also, I think you're better off writing `((Int, Int), V)` than `(Int, Int V)` because tuple2 is specialized and it won't write the type token twice.

johnynek:
Thanks for sending this. A few minor comments.

alekseyig:
@johnynek,  Hello Oscar, what is your opinion on that PR, will you merge it into master? Thank you.

chill-algebird/src/test/scala/com/twitter/chill/algebird/AlgebirdSerializersSpec.scala
```diff
@@ -16,15 +16,22 @@ limitations under the License.
...
 class AlgebirdSerializersSpec extends WordSpec with Matchers {
+  implicit def string2Bytes(i: String) = i.toCharArray.map(_.toByte)
+  implicit val skmMonoid = SketchMap.monoid[String, Long](SketchMapParams[String](1, 0.01, 1e-3, 10))
+
   val kryo = {
     val inst = () => {
       val newK = (new ScalaKryoInstantiator).newKryo
       newK.setReferences(false) // typical in production environment (scalding, spark)
+      newK.register(classOf[com.twitter.algebird.SketchMap[_, _]], new SketchMapSerializer)
       (new AlgebirdRegistrar).apply(newK)
```
johnynek: we should add this to the AlgebirdRegistrar. Sorry I didn't notice that before.

johnynek:
looks good. Sorry I overlooked it. I'll merge with one minor change (add this to AlgebirdRegistrar).

chill-algebird/src/test/scala/com/twitter/chill/algebird/AlgebirdSerializersSpec.scala
alekseyig: Yeah, I have tried it before and, sorry, I do not know how to do that generic way. The `SketchMapSerializer` requires the implicit `SketchMapMonoid`, which need to be specified before, but to do that I need to create one with very specific `SketchMapParams` as in 
` implicit val skmMonoid = SketchMap.monoid[String, Long](SketchMapParams[String](1, 0.01, 1e-3, 10))`. Can you take a look and help? Thank you.

chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
```diff
@@ -95,3 +86,42 @@ class HLLMonoidSerializer extends KSerializer[HyperLogLogMonoid] {
     hllMonoids.getOrElseUpdate(bits, new HyperLogLogMonoid(bits))
   }
 }
+
+class SketchMapSerializer[K, V](implicit skmMonoid: SketchMapMonoid[K, V], valueMonoid: Monoid[V])
```
johnynek: okay, I had somehow overlooked the issue these cause. Sorry. Ideally there are no parameters to a kryo serializer, but that's not easy to do given how AdaptiveMatrix holds on to the `Monoid[V]` which was probably a design flaw.

Could we get by here at least only holding the valueMonoid? I'm not sure why you need skmMonoid too, since it looks like you are writing the rows and columns from the SketchMap instance itself.

chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
```diff
@@ -95,3 +86,42 @@ class HLLMonoidSerializer extends KSerializer[HyperLogLogMonoid] {
...
+    val rowsOrig = in.readInt(true)
+    val colsOrig = in.readInt(true)
+    val values = kryo.readClassAndObject(in).asInstanceOf[Seq[((Int, Int), V)]]
+    val totalValue = kryo.readClassAndObject(in).asInstanceOf[V]
+    val heavyHitterKeys = kryo.readClassAndObject(in).asInstanceOf[List[K]]
+
+    val rows = if (rowsOrig == 0) skmMonoid.params.depth else rowsOrig
+    val cols = if (colsOrig == 0) skmMonoid.params.width else colsOrig
+
+    val zero = AdaptiveMatrix.fill[V](rows, cols)(valueMonoid.zero)
```
johnynek: there is `AdaptiveMatrix.zero(rows, cols)` for this case.

chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
```diff
@@ -95,3 +86,42 @@ class HLLMonoidSerializer extends KSerializer[HyperLogLogMonoid] {
...
+  }
+
+  def read(kryo: Kryo, in: Input, cls: Class[SketchMap[K, V]]): SketchMap[K, V] = {
+    val rowsOrig = in.readInt(true)
+    val colsOrig = in.readInt(true)
+    val values = kryo.readClassAndObject(in).asInstanceOf[Seq[((Int, Int), V)]]
+    val totalValue = kryo.readClassAndObject(in).asInstanceOf[V]
+    val heavyHitterKeys = kryo.readClassAndObject(in).asInstanceOf[List[K]]
+
+    val rows = if (rowsOrig == 0) skmMonoid.params.depth else rowsOrig
```
johnynek: can you explain these two lines of code?

chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
alekseyig: Yes, i can make this change.

chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
alekseyig: I am not 100% sure what lines you refer to, but would assume that 
` val rows = if (rowsOrig == 0) skmMonoid.params.depth else rowsOrig` and `val cols = if (colsOrig == 0) skmMonoid.params.width else colsOrig`.
 As I was saying that we wrote this serializer when we were using algebird 0.4 and in that version the implementation of `SketchMap` had a bug/feature that when you serialize empty `SketchMap` the `val rows = skm.valuesTable.rows` and `val cols = skm.valuesTable.cols` (see "write" method) would be set to 0 and to the size of the underneath supporting AdaptiveMatrix. Since this PR is for version of chill which is using algebird 0.10 these lines can be removed (I would prefer to still have them, so we do not need to recalculate our old SkatchMaps, but if it is a deal breaker, sure I can remove them).

chill-algebird/src/main/scala/com/twitter/chill/algebird/AlgebirdSerializers.scala
alekseyig: The reason we have skmMonoid is to support old serialized versions of SketchMap produced with old version of algebird 0.4, but can be eliminated (please see the other comment about that for more info).
You know, I have tried to remove this implicit SketchMapMonoid, but nor luck, I got into the deep loophole of implicit resolution order that I do not have a solution for. I would seek for your help here, please. Thank you.

alekseyig:
@johnynek, Hello Oscar, are you still interested in this pr or you decided to reject it after all this discussion? Thank you.

PR #236, opened 2015-07-28
Adds Scrooge 2.11 support now that the artifact is out, sets up for c…

ianoc:
…ross publishing


johnynek:
looks like the jar was not found.

ianoc:
@johnynek sorry, i dunno what I did there. Right jar version in place now. Looks green

PR #238, opened 2015-08-17
Adds two Injection's to lift T into an Externalizer of T, and codec f…

ianoc:
…or Externalizer[T]

### DIFF
--- a/chill-bijection/src/main/scala/com/twitter/chill/ExternalizerCodec.scala
+++ b/chill-bijection/src/main/scala/com/twitter/chill/ExternalizerCodec.scala
@@ -28,6 +28,9 @@ import scala.util.Try
 
 /**
  * A Injection to serialize an externalizer to bytes
+ *
+ * This method isn't safe since calling get on an Externalizer[T] could still throw despite the inversion
+ * here having worked as advertized. Since it is at the point of the get we unpack the inner box.
  */
 object ExternalizerCodec {
   implicit def apply[T]: ExternalizerCodec[T] = new ExternalizerCodec[T]
--- a/chill-bijection/src/main/scala/com/twitter/chill/ExternalizerCodec.scala
+++ b/chill-bijection/src/main/scala/com/twitter/chill/ExternalizerCodec.scala
@@ -1,0 +1,48 @@
+/*
+Copyright 2012 Twitter, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


chill-bijection/src/main/scala/com/twitter/chill/ExternalizerCodec.scala
```diff
@@ -0,0 +1,48 @@
...
+}
+
+class ExternalizerCodec[T] extends Injection[Externalizer[T], Array[Byte]] {
+  def apply(extern: Externalizer[T]): Array[Byte] = {
+    val baos = new ByteArrayOutputStream()
+    val oos = new ObjectOutputStream(baos)
+    oos.writeObject(extern)
+    baos.toByteArray
+  }
+  def invert(bytes: Array[Byte]): Try[Externalizer[T]] = Try {
```
johnynek: still not totally safe, because if you deserialize `Externalizer[String]` or `Externalizer[Int]` both will not throw until you call .get and it fails to cast to `String` or `Int`.

We could solve one layer deep, but then it could also fail for collections: `Externalizer[List[T]]`.

So, I'd rather just add a comment that this is unsafe and can result in class cast exceptions at runtime.

chill-bijection/src/main/scala/com/twitter/chill/ExternalizerInjection.scala
```diff
@@ -0,0 +1,43 @@
...
+}
+
+import com.twitter.bijection.Injection
+import scala.util.Try
+
+/**
+ * An Injection to lift a type T into an Externalizer
+ */
+object ExternalizerInjection {
+  implicit def apply[T]: ExternalizerInjection[T] = new ExternalizerInjection[T]
```
johnynek: apply is not a great name for an implicit do name shadowing causing issues there, I believe. Maybe call the implicit something else and have it call apply?

chill-bijection/src/main/scala/com/twitter/chill/ExternalizerCodec.scala
ianoc: Ok cool, seems reasonable

chill-bijection/src/main/scala/com/twitter/chill/ExternalizerInjection.scala
ianoc: orly? didn't know of such a thing. Will change

PR #243, opened 2015-10-22
If the inner Instantiator is cached we should just cache the kayo and…

ianoc:
… KryoPool also

Some of the WS looks off here as the file seems to have a mix of tabs and spaces. I'll follow up this PR with one replacing all tabs with spaces.


chill-hadoop/src/main/java/com/twitter/chill/hadoop/KryoSerialization.java
```diff
@@ -38,6 +38,27 @@
...
+     * so here we piggy back on that to avoid generating new kryo's or kryo pools
+     */
+    public static synchronized void resetOrUpdateFromCache(KryoSerialization instance, KryoInstantiator kryoInst){
+      if(kryoInst != cachedKryoInst) {
+        cachedTestKryo = kryoInst.newKryo();
+        cachedPool = KryoPool.withByteArrayOutputStream(MAX_CACHED_KRYO, kryoInst);
+        cachedKryoInst = kryoInst;
+      }
+      instance.kryoPool = cachedPool;
+      instance.testKryo = cachedTestKryo;
```
johnynek: I'm worried about the thread-safety of testKryo. We could be sharing it across threads, and there is no guarantee it is thread-safe.

Could we either borrow from the KryoPool to test or use a lock to guard access?

ianoc:
Updated this one to use the kryo pool for the tests

johnynek:
this looks good. Have you tried a scalding E2E run with this?

This is binary compatible, right? Can republish.

ianoc:
Would need to back port -- chill's dependency on algebird makes it not binary compatible at twitter currently. And yes I think so, can do a mima to be sure before publish anyway

johnynek:
@ianoc is it really not compatible? I think the dependency is pretty weak. Excluding algebird here should work. (Actually algebird should be provided).

ianoc:
It should be... but finding/hitting these has been pretty time consuming. Backporting is faster/easier in the end. (Scala compiler can detect some class you refer to in a jar isn't valid against a transitive dependency anymore). So I've backported changes... will see if I can test them tonight

ianoc:
Twitter internal e2e's passed on this change fyi

johnynek:
Awesome. Merge here and in a chill08 branch?

PR #244, opened 2015-11-13
Upgrade to kryo-2.22

steveloughran:
This is a patch which moves Chill up to Kryo 2-22; so giving chill and scala a version of scala consistent with hive 1.2.x

see https://github.com/twitter/chill/pull/230 for related discussions on a move to kryo 3, which is probably a more traumatic event

### DIFF
--- a/chill-avro/src/test/java/com/twitter/chill/avro/AvroSerializerJavaTest.java
+++ b/chill-avro/src/test/java/com/twitter/chill/avro/AvroSerializerJavaTest.java
@@ -11,7 +11,7 @@ import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecordBuilder;
 import org.junit.Before;
 import org.junit.Test;
-import org.objenesis.strategy.StdInstantiatorStrategy;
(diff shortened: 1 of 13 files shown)


ianoc:
Multiple large source/binary incompatible switches are a huge friction for people. I don't see why we would try do several given all of them will require synchronization in lots of places for people

steveloughran:
I do fully understand that; with kryo's repackaging being a particular troublespot: you can't drop in the new JAR, you have to edit source.

But 
1. moving chill -> Kryo 2.22 doesn't need co-ordination with Hive; it is already there.
2. Hive has to play games so that its import of spark doesn't create transitive kryo problems.
3. Spark has to play games so that its import of hadoop doesn't create transitive kryo problems.
4. yes, that is a loop in the build.
5. I don't see Spark adopting Kryo 3 until a 2.0 release, which means this problem will continue to exist

Hives upgrade to 2.22 fixed a lot of problems; I don't know if Kryo 3 will help fix some other issues, or introduce other surprises: they haven't shipped with Kryo 3 so won't know. Kryo 2.22 is not a change they'd roll back.

Upgrading chill to 2.22 would impact scala apps downstream, your own & spark; once done it would result in a consistent version with Hive (at least currently).

PR #245, opened 2015-11-16
Kryo 3.0.3 upgrade

ianoc:
Forked from https://github.com/twitter/chill/pull/230

Remove the java version limits, we support from 1.6+ still


chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
```diff
@@ -117,8 +117,8 @@ object Instantiators {
   }
 
   // Use call by name:
-  def forClass(t: Class[_])(fn: () => Any): ObjectInstantiator =
-    new ObjectInstantiator {
+  def forClass(t: Class[_])(fn: () => Any): ObjectInstantiator[AnyRef] =
```
johnynek: shouldn't this be:
```scala
def forClass[T](t: Class[T])(fn: () => Any): ObjectInstantiator[T] =
    new ObjectInstantiator[T] {
       override def newInstance() = {
         try { t.cast(fn()) }
```

chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
```diff
@@ -130,7 +130,7 @@ object Instantiators {
     }
 
   // This one tries reflectasm, which is a fast way of constructing an object
-  def reflectAsm(t: Class[_]): Either[Throwable, ObjectInstantiator] = {
+  def reflectAsm(t: Class[_]): Either[Throwable, ObjectInstantiator[AnyRef]] = {
```
johnynek: also can we add the `T` parameter here?

chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
```diff
@@ -154,7 +154,7 @@ object Instantiators {
     }
   }
 
-  def normalJava(t: Class[_]): Either[Throwable, ObjectInstantiator] = {
+  def normalJava(t: Class[_]): Either[Throwable, ObjectInstantiator[AnyRef]] = {
```
johnynek: add `T`

johnynek:
looks good to me.

Ian, what are the issues with using this with summingbird, which hits storm? Have they shaded their kryo?

ianoc:
No, not shaded, at twitter i've talked to the heron folks about it. External storm users are going to need carbonite upgraded iirc.

johnynek:
but will our spark tests fail? I'd really hate to break external users of summingbird.

ianoc:
spark tests? Pretty much every downstream test will fail until fixed for this sort of change. Its pretty much unavoidable though or we can never upgrade kryo :/ 

Our storm tests in summingbird are the most at risk unit tests to fail that are tricky to fix that I know of. Spark doesn't require kryo usage so until spark upgrades our tests can turn off those paths

johnynek:
yeah, I meant summingbird tests. I guess we have to start moving up the stack.

steveloughran:
FWIW, here's the documented experience of the hive move [HIVE-12175](https://issues.apache.org/jira/browse/HIVE-12175)

ozawa-hi:
I'm trying to get kyro3 upgraded in upcoming Spark2.0. Is it possible to get this in maven repository so I can update the pom.xml file in Spark? I'm already testing in local env but would like to create a branch at Spark for this.
https://issues.apache.org/jira/browse/SPARK-11416

johnynek:
@ozawa-hi +1 the time it now to go forward.

chill-hadoop/src/main/java/com/twitter/chill/hadoop/KryoDeserializer.java
```diff
@@ -46,7 +46,7 @@ public void open(InputStream in) throws IOException {
 
     public Object deserialize(Object o) throws IOException {
         // TODO, we could share these buffers if we see that alloc is bottlenecking
-        byte[] bytes = new byte[inputStream.readInt()];
+        byte[] bytes = new byte[(int) Varint.readUnsignedVarLong(inputStream)];
```
johnynek: is there no int? should we throw here if the long > Int.MaxValue

johnynek: Oh, the code is below. Why not unsigned varint?

chill-hadoop/src/main/java/com/twitter/chill/hadoop/Varint.java
```diff
@@ -0,0 +1,168 @@
...
+        // This undoes the trick in writeSignedVarInt()
+        int temp = (((raw << 31) >> 31) ^ raw) >> 1;
+        // This extra step lets us deal with the largest signed values by treating
+        // negative results from read unsigned methods as like unsigned values.
+        // Must re-flip the top bit if the original read value had it set.
+        return temp ^ (raw & (1 << 31));
+    }
+
+    /**
+     * @throws IllegalArgumentException if variable-length value does not terminate
```
johnynek: this is a lie, right?

chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
```diff
@@ -101,9 +103,9 @@ class KryoBase extends Kryo {
 
 object Instantiators {
   // Go through the list and use the first that works
-  def newOrElse(cls: Class[_],
-    it: TraversableOnce[Class[_] => Either[Throwable, ObjectInstantiator]],
-    elsefn: => ObjectInstantiator): ObjectInstantiator = {
+  def newOrElse[T](cls: Class[T],
+    it: TraversableOnce[Class[T] => Either[Throwable, ObjectInstantiator[T]]],
```
johnynek: `Either[Throwable, A]` should be `Try[A]`. This code predates scala Try.

ozawa-hi:
Spark upgraded their Scala on Master branch to 2.11. Is it possible to change scalaVersion := "2.11.7", in Build.scala?

ianoc:
We publish all these projects for both 2.10 and 2.11, no reason why we would change the default build. Doesn't effect anything but chill itself

chill-hadoop/src/main/java/com/twitter/chill/hadoop/Varint.java
ianoc: Looks like one(this code isn't mine), add test, or kill comment? 

```
        while (((b = in.readByte()) & 0x80) != 0 && i < 42) {
            value |= (b & 0x7F) << i;
            i += 7;
        }
        if(i == 42) { // Over read!
            throw IllegalArgumentException("Read more than 5 bytes of data, must be invalid Var int")
        }
```
Looks like it would fix this one

chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
ianoc: Will update file

chill-hadoop/src/main/java/com/twitter/chill/hadoop/KryoDeserializer.java
ianoc: I think this flowed out of the kryo api for bytes written returning a long... at the entry point to that i goto an int now, so everything else just ends up being integers

johnynek:
https://github.com/apache/storm/blob/master/pom.xml#L231

johnynek:
https://issues.apache.org/jira/browse/HIVE-11269?jql=text%20~%20%22kryo%22

what a pain we have without isolated classpaths.

johnynek:
https://github.com/sritchie/carbonite/blob/master/project.clj#L8 still in play:

https://github.com/apache/storm/blob/02a44c7fc1b7b3a1571b326fde7bcae13e1b5c8d/pom.xml#L210

actually, transitively, storm depends on chill 0.3.5:
https://github.com/sritchie/carbonite/blob/master/project.clj#L9

johnynek:
so, one way forward may be:

1. open issue with storm.
2. publish chill-kryo2 and chill as the old and new branches of this library.
3. in summingbird, make the summingbird-storm depend on chill-kryo2?

in any case, I think publishing either a new artifact or a new version is the way to unblock spark.

johnynek:
https://issues.apache.org/jira/browse/STORM-1537

ianoc:
Storm look down to do the update if we publish a new version based on your JIRA there. Should we just push ahead and do a bump? Scalding won't take it till 0.17 anyway, so we have some runway for storm to update everything else. (Summingbird will need to update to a newer storm at the same time). 

Once we merge this call it 0.8.0 or so?

johnynek:
@ianoc sounds good to me.

ianoc:
Ok awesome, sounds like a plan. Any other changes you'd like to this PR pre-merge?

The other active one can be applied to the 0.7.x series and 0.8.x series presuming all tests are good there, so i'll try put it against both.

PR #247, opened 2015-12-26
Use currentThread classloader

### LINKED ISSUE #246: We should always use correct classloader
see a manifestation here: http://themodernlife.github.io/scala/hadoop/hdfs/sclading/flink/streaming/realtime/2015/12/20/running-scalding-jobs-on-apache-flink/

also issue: #194

johnynek:
closes #246


ianoc:
Sigh jvm. would love if we could easily add tests for some of this stuff. But seems like obviously an improvement and something we need here.

PR #260, opened 2016-08-25
Serializers for unmodifiable Java collections.

chermenin:
Added serializers for collections created via
Collections.unmodifiable...() and Collections.empty...() methods.


chill-java/src/main/java/com/twitter/chill/java/UnmodifiableCollectionSerializer.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * A kryo {@link Serializer} for unmodifiable collections created via {@link Collections#unmodifiableCollection(Collection)}.
+ * <p>
+ * Note: This serializer does not support cyclic references, so if one of the objects
+ * gets set the list as attribute this might cause an error during deserialization.
+ * </p>
+ *
+ * @author <a href="mailto:alex@chermenin.ru">Alex Chermenin</a>
+ */
+public class UnmodifiableCollectionSerializer extends Serializer<Collection<?>> {
```
johnynek: these are all basically the same. Can we share the code? Something like:

``` java
public abstract class UnmodifiableJavaCollectionSerializer<T, U> extends Serializer<T> {
    protected U getInner(T t);
    protected T build(U u);

    @Override
    public Collection<?> read(Kryo kryo, Input input, Class<T> type) {
        try {
            U c = (U) kryo.readClassAndObject(input);
            return build(c);
        } catch (Exception e) {
            throw new RuntimeException(e);
        }
    }

    @Override
    public void write(Kryo kryo, Output output, T object) {
        try {
            U c = getInner(object);
            kryo.writeClassAndObject(output, c);
        } catch (RuntimeException e) {
            // Don't eat and wrap RuntimeExceptions because the ObjectBuffer.write...
            // handles SerializationException specifically (resizing the buffer)...
            throw e;
        } catch (Exception e) {
            throw new RuntimeException(e);
        }
    }
}
```

Then we can just implement the code to read the inner and wrap for each of them.

Also, could you possibly set your editor to 2 space indent, which is the style we use?

chermenin:
Great thought! Thanks for review.

PR #271, opened 2016-12-11
Clean up build before releasing 2.12

sritchie:
This PR:

- Re-enables our MIMA checks
- Turns on codecov support
- Modifies chill-akka's implementation to get it ready for 2.4, when/if we upgrade
- turns on more modules for 2.12! (we can turn on Algebird's module after deploying on Monday.)
- turns on caching in the build
- turns on gitter notifications
- adds better badges to README.md
- upgrades sbt to 0.13.13, fixes deprecation warnings

/cc @isnotinvain @johnynek


sritchie:
Sooo 2.10 fails with the Akka upgrade, since Akka only publishes for two scala versions. @johnynek @isnotinvain, we can either

remove the akka upgrade, and keep chill-akka on 2.10/2.11 and an old version, or
think about dropping support for 2.10 in our builds. 

If it's not a direct problem for any of us I vote for dropping 2.10 support. 2.11 came out 2.5 years ago; this is plenty of time to upgrade, and 2.10 came out over a year before that (march 2013). I think supporting two major versions is absolutely fine, and would MUCH rather support modern users of Akka on chill.

Thoughts?

sritchie:
Turns out my akka fixes are compatible with the version we were using before - I've rolled back the version but kept those fixes.

sritchie:
Another note - if we're willing to upgrade to scrooge 4.12, we can get that module as well.

johnynek:
This looks good. The issue is that we also want 2.12 for the 0.7 branch, I think. Chill is very complex because different branchs run different Kryo versions, but some users (like us) are nervous what we have stored some Kryo serialized instances that we can deserialize with a later version.

So, I'm +1 on this but we probably need to follow up with another PR against the 0.7 branch (and maybe we need to make a 0.9 branch too to use the latest Kryo).

johnynek (approved):

ngocdaothanh:
Can you release a new version so that I can use it in Scala 2.12 projects?

sritchie:
Hey @ngocdaothanh, @johnynek and I are trying to find some time to do the release this week. Sorry for the delay! It's been tough to fit it in the cracks between other work this week.

Issue #266, opened 2016-11-04
Add Scala 2.12 support


LukaszByczynski:
The title says it all

jodersky:
this is happening in #253

ritschwumm:
which is merged now - how about a release?

ritschwumm:
pretty please?

sritchie:
@ritschwumm, @johnynek and I are going to take care of this on Monday. We cleaned up the build a bit with #271 - we'll release algebird first so we can get 2.12 support for chill-algebird as well. Thanks for the push!

fkowal:
any luck?

ritschwumm:
well, he didn't say _which_ monday ;)

johnynek:
okay... I'll work on this now.

It's quite complex because we have three branches 0.7, 0.8, and 0.9 each with different Kryos. We need to release for each version on each of those branches.

johnynek:
https://github.com/twitter/chill/releases/tag/v0.9.0

There is 0.9.0 with Kryo 4.0.0. Next is 0.8.2 with Scala 2.12.

fkowal:
maven might still be indexing
but http://search.maven.org/#search%7Cga%7C1%7Cchill-akka
i can't seem to find any chill-akka_2.12 artifacts

johnynek:
https://github.com/twitter/chill/pull/272

here is the 0.8 branch.

johnynek:
@fkowal looks like akka was explicitly excluded: https://github.com/twitter/chill/pull/272/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR156 maybe akka was not on 2.12 when the PR was made?

fkowal:
sure looks like it, 2.12 support in akka starts from version 2.4.12

johnynek:
can you make a PR to fix the build for akka? I don't use and have never used akka in production, so I don't really know what the story is.

Generally, we want the lowest version available that will cover the most commonly used versions in production.

Akka should probably be a provided dependency, and the person using chill-akka should set the version they need (so we just need to have some version that has the same basic API for the parts we use).

sritchie-stripe:
I actually did this a few weeks ago:

https://github.com/twitter/chill/tree/sritchie/akka_upgrades

@fkowal, would you mind taking the fixes from that branch and sending a new PR with the version of Akka that you prefer?

fkowal:
on it

fkowal:
[...]

PR #273, opened 2017-01-03
chill-akka support scala 2.12 #266

fkowal:


fkowal:
hmm one of travis build failed with a strange error.
sbt +test runs successfuly on my machine

can i retrigger a travis job ?

johnynek:
That is not a strange error, that is the error when there is a mismatch between the compiled class versions and the jdk you are compiling on. In this case, I guess you need java8 for the 2.11 branch, so you need to update that in .travis.yml.

johnynek:
@fkowal if you fix this PR to use oraclejdk8 on 2.11 tests in `.travis.yml` I'll backport this to the 0.7 and 0.8 branches so we can have things in sync there.

fkowal:
that helped

johnynek:
Thanks!

mushtaq:
@johnynek I still do not see the chill-akka_2.12 artifact in mavencentral.

johnynek:
@mushtaq should be there shortly. Was published today. cc #266

mushtaq:
@johnynek  Thanks!

PR #277, opened 2017-01-10
Build the docs on travis

johnynek:
publishing needs to build docs. Looks like some javadocs are currently broken. This adds building docs to travis (which I think will fail for 2.10) and then I'll get this green.

Need to also do this for 0.7.x and develop.


johnynek:
@sritchie how does this look? Fix the docs and build them in CI so they don't get broken again.

sritchie:
👍

PR #290, opened 2017-03-24
Fix extra allocation + copy in SerDeState

jcdavis:
There appears to be no coverage of `writeOutputTo`, so I added something simple to explicitly test it.


johnynek (approved):
Thanks a lot!

PR #296, opened 2017-08-31
Use relatively safe max size of java array

ravwojdyla:
More info:
 * https://github.com/EsotericSoftware/kryo/pull/516
 * https://stackoverflow.com/questions/3038392/do-java-arrays-have-a-maximum-size


johnynek:
Thank you!

PR #301, opened 2018-01-14
Handle Kryo copying of scala immutable objects

MThomassen:
current scala serializer classes flag their immutability by overriding the com.esotericsoftware.kryo.Serializer::isImmutable member.

However, on Kryo copy activity, private isImmutable field of the serializer is read instead of this public member, therefore, this overrides intention is ignored.

New scala serializers for immutable objects set immutability through the Kryo.Serializer::setImmutable member so private serializer fields will be correctly set.

### DIFF
--- a/chill-scala/src/main/scala/com/twitter/chill/BitSetSerializer.scala
+++ b/chill-scala/src/main/scala/com/twitter/chill/BitSetSerializer.scala
@@ -19,6 +19,8 @@ package com.twitter.chill
 import scala.collection.immutable.BitSet
 
 class BitSetSerializer extends KSerializer[BitSet] {
+  setImmutable(true)
+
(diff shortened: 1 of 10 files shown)


johnynek:
Thanks for the PR.

I’m not sure it is correct to say generic items are immutable. They are only immutable if what they point to is immutable.

For instance: `Some(Array(1, 2, 3))` is not immutable. Is this a problem?

I think we may have made this error in the past, but it seems this PR increases the cases where it can be made. What do you think?

MThomassen:
@johnynek I think you're right; I've reverted the immutable-flag to 'false' for these generic objects

I've also bungled the CLA signing (pushed from a PC without git config), could someone more knowledgeable than me fix the signing?

Issue #227, opened 2015-04-14
Missing scala.collection.:: in registration


rcoh:
The following will fail if registration is required:

```
serializer.serialize(List(1,2,3))
```

This is because the collections list is missing `::` and only has the empty list:
```     
 // List is a sealed class, so there are only two subclasses:
      .forTraversableSubclass(List.empty[Any])
```

The fix is just to add list, but it needs to be added to the end or all of the registration numbers will change and it will break clients.

ianoc:
We don't expect Chill serialization to be stable across jvm launches/recompiles so it should be fine to add it in to the natural place in the constructor. Would happily take a PR

peter-empen:
Could you please elaborate on the above statement "We don't expect Chill serialization to be stable across jvm launches/recompiles"?
We do expect registration numbers to be stable. How do you expect users to upgrade Chill otherwise?
Btw., is this issue still valid?

ianoc:
Chill was never designed to be used as a long term serialization format, since the underlying kryo itself wasn't really designed for this. Kryo has become a bit more stable over time in this regard but i'd still not recommend ever using kryo/chill as a long term format. 

The common place this is used as a magic/auto-serializer between hosts in a distributed system. Its used in Scalding/Spark/Summingbird for this, Storm uses kryo too iirc.

peter-empen:
Our use case spans Akka Distributed _and_ Event Stores.
We are confident that, once registry numbers are dealt with care, we do achieve kinda long-term format. For this purpose, we certainly work with _RegistrationRequired_.

ianoc:
Chill has no tests requiring these numbers are constant from build to build. So i think you can propose a different tack and all the tests here. But barring that if the tests don't require it I don't think you should rely on this. Using something like protobuf/thrift/etc.. that all aim for backwards compatibility/LTS. 

(Side note where I am now an online event store was done using chill/kryo for serialization persistance, and its been a total pain since we can't easily upgrade kryo in anything touching its classpath.)

peter-empen:
We do have these tests already.
Yeh, we did realize upgrading Kryo over major releases won't be possible. Why do you think we will be hit by this restriction seriously? What about using current Kryo with current feature set for some years?

ianoc:
[...]

PR #302, opened 2018-01-31
Pre-registering certain standard classes to better support long term persistence scenarios

gdiet:
This PR is about registering standard scala classes that are likely to be used when long term persisting data, for example in event stores:

When using kryo/chill for long term storage, it is advisable to run Kryo with setRegistrationRequired(true). However, a number of standard scala classes that are useful in such scenarios are not pre-registered in chill, most of them located in scala.collections.immutable. This has puzzled others as well, see for example the chill issue #227 which is fixed by this PR.

Some of the classes I have added to the standard registration in this pull request:

scala.None
scala.collection.immutable.::
scala.collection.immutable.Nil
scala.collection.immutable.ListSet.EmptyListSet
scala.collection.immutable.Range
scala.collection.mutable.WrappedArray.ofRef
int[]

To better support the long term persistence scenario, I have added the RegistrationIdsSpec, which will fail if the pre-registered IDs change, because changing them should be avoided if possible to ensure backwards compatibility. Or, if changing pre-registered IDs is really inevitable, a compatibility note should be added to the release notes.

If this PR is accepted, I plan to complement it with two more PRs: One PR for the chill documentation with notes on how kryo/chill can be used for long term persisting events (lessons learned). The other PR would be another spec ensuring that the binary representation of serialized classes is not changed unintendedly.


gdiet:
Yes, Travis CI build failed, but :) this is because travis uses scala 2.10.6, and with scala 2.10.7 the tests run green (at least locally). So I propose that the Travis CI is reconfigured to use scala 2.10.7.

johnynek:
@gdiet can you bump to modern scalas here:

https://github.com/twitter/chill/blob/develop/.travis.yml

then you should be green?

johnynek:
Thanks for this PR, by the way!

chill-scala/src/main/scala/com/twitter/chill/ScalaKryoInstantiator.scala
```diff
@@ -121,15 +98,32 @@ class ScalaCollectionsRegistrar extends IKryoRegistrar {
      */
     newK
       // wrapper array is abstract
+      .registerClasses(Seq(
```
johnynek: can you add all new items at the end of the list? Otherwise, I think we could change the tokens and some people have (ill-advisedly) persisted kryo data into long term storage.

chill-scala/src/test/scala/com/twitter/chill/RegistrationIdsSpec.scala
```diff
@@ -0,0 +1,179 @@
+package com.twitter.chill
+
+import org.scalatest.{ Matchers, WordSpec }
+
+class RegistrationIdsSpec extends WordSpec with Matchers {
+  """
+    |Projects using chill to long term persist serialized data (for example in event
```
johnynek, line 7: so this test protects the future, but not people in the past right? They all get broken right?

Could we work around that? Maybe add a new registrar that people can opt in to adding?

johnynek (commented):
This looks like a great improvement. I appreciate the tests. Thank you.

chill-scala/src/main/scala/com/twitter/chill/ScalaKryoInstantiator.scala
gdiet: yes... I will follow your next proposal and put all new items into a separate registrar.

chill-scala/src/test/scala/com/twitter/chill/RegistrationIdsSpec.scala
gdiet: Right, this would break compatibility for projects that have already used chill for long term storage. Defining a separate registrar is OK for me (although the order of entries in the registrar will be less orderly that way).

I'm thinking along these lines currently:
* Rename AllScalaRegistrar -> AllScalaRegistrar_0_9_2 for those who need the old registrations.
* Create a new AllScalaRegistrar that first registers everything from AllScalaRegistrar_0_9_2, then registers all the new items from this PR.

Does this sound OK?

PR #303, opened 2018-02-02
Pre-registering certain standard classes to better support long term persistence scenarios

gdiet:
_This PR replaces https://github.com/twitter/chill/pull/302, there have been problems caused by commits containing an incorrect committer's email address._

This PR is about registering standard scala classes that are likely to be used when long term persisting data, for example in event stores:

When using kryo/chill for long term storage, it is advisable to run Kryo with setRegistrationRequired(true). However, a number of standard scala classes that are useful in such scenarios are not pre-registered in chill, most of them located in scala.collections.immutable. This has puzzled others as well, see for example the chill issue #227 which is fixed by this PR.

Some of the classes I have added to the standard registration in this pull request:

scala.None
scala.collection.immutable.::
scala.collection.immutable.Nil
scala.collection.immutable.ListSet.EmptyListSet
scala.collection.immutable.Range
scala.collection.mutable.WrappedArray.ofRef
int[]

To better support the long term persistence scenario, I have added the RegistrationIdsSpec, which will fail if the pre-registered IDs change, because changing them should be avoided if possible to ensure backwards compatibility. Or, if changing pre-registered IDs is really inevitable, a compatibility note should be added to the release notes.

For projects that already depend on the class ID registration present in chill 0.9.2, there is a compatibility registrar `AllScalaRegistrar_0_9_2` defined.

If this PR is accepted, I plan to complement it with two more PRs: One PR for the chill documentation with notes on how kryo/chill can be used for long term persisting events (lessons learned). The other PR would be another spec ensuring that the binary representation of serialized classes is not changed unintendedly.


gdiet:
@johnynek thank you for your comments to https://github.com/twitter/chill/pull/302 - this PR replaces that one, and I have updated things to address the concerns you mentioned.

johnynek (approved):
This is great, thanks!

PR #304, opened 2018-03-26
Chill-Akka sample configuration

raboof:
Configuring Akka to use Kryo for all `scala.Product` classes may be problematic because many of our internal messages are also case classes, and are expected to be serialized with our internal serializer (e.g. `InitJoin` with `ClusterMessageSerializer`).

Since our marker interfaces do not inherit `scala.Product` but do inherit `java.io.Serializable`, the latter is safer to reconfigure (because of the 'prefer the most specific serializer' semantics). The additional benefit is that this unlocks disabling Java serialization entirely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,12 +141,27 @@ To use, add a key to your config like:
     }
 ```
 
-Then for the super-classes of all your message types, for instance, scala.Product, write:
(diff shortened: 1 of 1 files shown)


johnynek (approved):
Thank you so much!

PR #305, opened 2018-03-30
add QTreeSerializer

nevillelyh:
Default kryo serialization for `QTree` is not deterministic due to other fields in the class, and this causes problem in some frameworks like Apache Beam which detects mutation of input element via serialization.


johnynek (approved):
Thanks!

PR #313, opened 2018-06-15
fixed tests broken by PR #312

gdiet:
PR #312 registered the serializer for BigDecimal with ID 15, moving all serialization IDs from 15 on one up. This breaks compatibility for projects using chill for long term persistence - see com.twitter.chill.RegistrationIdsSpec which failed just for that reason.

In this commit, the registration of BigDecimal is moved to ID 144 (the end of the list) and the RegistrationIdsSpec is updated accordingly.


gdiet:
... it would be interesting to know why travis did not complain that tests fail ... ?

johnynek:
huh, this is worrisome.... thanks for sending the fix.

PR #317, opened 2018-08-13
Add class scala.collection.immutable.Queue$EmptyQueue$ to AllScalaRegistrar

gdiet:
The AllScalaRegistrar is supposed to register (among others) scala's basic collection classes. This PR adds registration for class scala.collection.immutable.Queue$EmptyQueue$.

When looking into the details, note that scala.collection.immutable.Queue was already registered in AllScalaRegistrar, and that Queue() is not the same object as Queue.empty[Any], although they compare equal.

To avoid breaking compatibility in products using chill for long term persistence, Queue.empty[Any] has been registered **at the end** of the list.


johnynek:
thank you!

PR #318, opened 2018-08-27
Pre-register Map.filterKeys, Map.mapValues, Map.keySet in the ScalaKryoInstantiator

gdiet:
immutable.Map's filterKeys, mapValues, and keySet methods create objects whose classes are not yet registered by ScalaKryoInstantiator. This PR adds these classes, so they are pre-registered and can easily be used together with setRegistrationRequired(true).

This is another addition for https://github.com/twitter/chill/pull/303.


gdiet:
another problem with committer settings - will create a follow-up PR (hopefully) fixing this.

PR #319, opened 2018-08-27
Pre-register Map.filterKeys, Map.mapValues, Map.keySet in the ScalaKryoInstantiator

gdiet:
immutable.Map's filterKeys, mapValues, and keySet methods create objects whose classes are not yet registered by ScalaKryoInstantiator. This PR adds these classes, so they are pre-registered and can easily be used together with setRegistrationRequired(true).

This is another addition for #303.

This PR supersedes #318 by (hopefully) fixing the committer settings.


nevillelyh (approved):
LGTM @johnynek any thoughts?

chill-scala/src/main/scala/com/twitter/chill/ScalaKryoInstantiator.scala
```diff
@@ -252,5 +252,8 @@ class AllScalaRegistrar extends IKryoRegistrar {
     k.forClass[scala.runtime.VolatileByteRef](new VolatileByteRefSerializer)
     k.forClass[BigDecimal](new BigDecimalSerializer)
     k.register(Queue.empty[Any].getClass)
+    k.register(Map(1 -> 2).filterKeys(_ != 2).getClass)
+    k.register(Map(1 -> 2).mapValues(_ + 1).getClass)
+    k.register(Map(1 -> 2).keySet.getClass)
```
johnynek: I think we should use `TraversableSerializer` here. I think `.forConcreteTraversableClass` will do the job.

gdiet:
Using .forConcreteTraversableClass makes sense - here we go.

gdiet:
@johnynek just a "ping" in case you have not seen my last addition to the PR using `.forConcreteTraversableClass` for the registrations.

johnynek:
sorry for the delay!

Issue #322, opened 2018-11-13
Serializing and deserializing case classes with KryoPool in scala


tarunk2:
```
val instantiator = new ScalaKryoInstantiator
    instantiator.setRegistrationRequired(false)
    val kryo = KryoPool.withByteArrayOutputStream(10, instantiator)

    "serialize and deserialize properly" in {
      val bytes = kryo.toBytesWithClass(Bar(1, "abc"))
      println(kryo.fromBytes(bytes, classOf[Bar]))
    }
```

```
case class Bar(a: Int, b: String)

```

It is failing with 
```
Caused by: java.lang.IndexOutOfBoundsException: Index: 97, Size: 1
	at java.util.ArrayList.rangeCheck(ArrayList.java:653)
	at java.util.ArrayList.get(ArrayList.java:429)
	at com.esotericsoftware.kryo.util.MapReferenceResolver.getReadObject(MapReferenceResolver.java:42)
	at com.esotericsoftware.kryo.Kryo.readReferenceOrNull(Kryo.java:805)
	at com.esotericsoftware.kryo.Kryo.readObjectOrNull(Kryo.java:728)
	at com.esotericsoftware.kryo.serializers.ObjectField.read(ObjectField.java:113)
```

Any ideas?

johnynek:
I think you are serializing inconsistently. If you use `fromBytes` and pass a `class` it assumes the class was *not* written to the stream I believe:

If you use fromBytes with a class, you should use toBytesWithoutClass:
https://github.com/twitter/chill/blob/0919984ec3aeb320ff522911c726425e9f18ea41/chill-java/src/main/java/com/twitter/chill/KryoPool.java#L124

What version of Kyro and Chill are you using?

We have many tests. Can you reproduce this error in a PR to this repo?

Is this an inner class of another class? Scala includes references to outer scope that could be calsuing an issue.

tarunk2:
Yep, that was it. Thanks! I was following README. Fixed the README in https://github.com/twitter/chill/pull/323

PR #323, opened 2018-11-13
fixing README - toBytesWithoutClass should be used with fromBytes

tarunk2:


johnynek:
I think the README is correct as is. Note the README does not pass a class argument unlike the example in #322 do you agree?

tarunk2:
You are right, my bad!

PR #326, opened 2019-01-30
Includes java regex pattern flags to serialization.

jcroig:
Pattern flags are ignored so in case a pattern with say case insensitive flags is serialised it will be deserialised in case sensitive mode which is indeed a different pattern.

### DIFF
--- a/chill-java/src/main/java/com/twitter/chill/java/RegexSerializer.java
+++ b/chill-java/src/main/java/com/twitter/chill/java/RegexSerializer.java
@@ -35,10 +35,11 @@ public class RegexSerializer extends Serializer<Pattern> {
     @Override
     public void write(Kryo kryo, Output output, Pattern pattern) {
         output.writeString(pattern.pattern());
+        output.writeInt(pattern.flags(), true);
     }
 
     @Override
     public Pattern read(Kryo kryo, Input input, Class<Pattern> patternClass) {
-        return Pattern.compile(input.readString());
+        return Pattern.compile(input.readString(), input.readInt(true));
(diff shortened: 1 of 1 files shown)


chill-java/src/main/java/com/twitter/chill/java/RegexSerializer.java
```diff
@@ -35,10 +35,11 @@ static public IKryoRegistrar registrar() {
...
     public void write(Kryo kryo, Output output, Pattern pattern) {
         output.writeString(pattern.pattern());
+        output.writeInt(pattern.flags(), true);
     }
 
     @Override
     public Pattern read(Kryo kryo, Input input, Class<Pattern> patternClass) {
-        return Pattern.compile(input.readString());
+        return Pattern.compile(input.readString(), input.readInt(true));
     }
```
johnynek, line 44: sorry I left this sit so long, but it would be nice to add a test that shows the problem.

nevillelyh:
@johnynek Just realized that this actually causes a regression. I didn't notice because CI was turned off. Should we revert it, or update the test and make this a breaking change? 
https://github.com/twitter/chill/pull/330#issuecomment-536698169

nevillelyh:
@jcroig I had to revert this as part of #330 since it causes a regression. Not sure how we can do it in a backwards compatible way though.

jcroig:
But don't you think this is a bug @nevillelyh? I mean the pattern is not the same before and after serialisation which should be, do you agree? What kind of regression is this creating is it breaking other tests or something?

PR #330, opened 2019-09-16
modernize build and update versions

nevillelyh:
Modernize the build a bit and prepare for 2.13 cross build.
Some tests are failing due to changes in tuple serializers, etc. Also Travis is not building?

- https://github.com/twitter/chill/pull/330/commits/810b30a40605a829676fd4f2cb4c31f84f6d2975 was needed after bumping 2.12 to 2.12.10. But I can save them for a separate PR.


chill-scala/src/main/scala/com/twitter/chill/TupleSerializers.scala
```diff
@@ -699,177 +720,36 @@ class Tuple22Serializer[A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S,
       kser.readClassAndObject(in).asInstanceOf[S],
       kser.readClassAndObject(in).asInstanceOf[T],
       kser.readClassAndObject(in).asInstanceOf[U],
-      kser.readClassAndObject(in).asInstanceOf[V])
+      kser.readClassAndObject(in).asInstanceOf[V]
+    )
   }
 }
-
-class Tuple1LongSerializer extends KSerializer[Tuple1$mcJ$sp] with Serializable {
```
johnynek: wait, removing these will be a performance regression I think (and break people who took their chances and stored these to disk. This set of classes with one type tag tell you both of the tuple types, which will save you from having to write them.

johnynek (commented):
I'd really like to find a way not to remove the specialized tuple serializers. Can we make a smaller change that doesn't change those? merge that, then address this question?

chill-scala/src/main/scala/com/twitter/chill/TupleSerializers.scala
nevillelyh: Yeah seems these specialized classes were renamed in 2.12.9. I reverted them back.
However those using chill with scala >= 2.12.9 will still experience the regression. Maybe we should remove them in 0.10 once we cross build 2.13?

nevillelyh:
`sbt test` passes for all 3 scala branches for me locally. Any idea why Travis isn't running?
https://travis-ci.org/twitter/chill says "This is not an active repository".

chill-scala/src/main/scala/com/twitter/chill/TupleSerializers.scala
smarter: https://github.com/smarter/chill/commits/scala213 handles the tuple serializer thing by rewriting it into Java: https://github.com/smarter/chill/commit/dd1fcd148ba2b8320d90f599cc9d204af13760f7. This should work just as well for 2.12.10, but maintaining binary compat would require some extra effort (which might not be useful in practice, I don't know if references to the tuple serializers are likely to appear in user code)

johnynek: @smarter I assume the classes themselves are still there, you can't can't directly reference them? Otherwise that would be a binary incompatible change in scala 2.12 itself.

I think breaking compatibility would be bad, since it is hard to reason about how these classpaths get set up. Storm, Scalding and Spark all depend on chill. It isn't obvious if they wind up depending on these serializers. Ideally we could make Java types with the same names that would be binary compatible. Maybe that isn't possible.

smarter: > Ideally we could make Java types with the same names that would be binary compatible. Maybe that isn't possible.

I think it's possible but will require some effort, it might be simpler to keep the current tuple serializers for 2.12-.

nevillelyh:
@johnynek do you have permission to re-enable Travis CI for this?
Also should we merge this (or a subset of it) and keep iterating on `TupleSerializer` and other fixes?

johnynek:
@nevillelyh I don't. @decause can maybe help out?

decause:
@nevillelyh & @johnynek 

Repository has been successfully reactivated on Travis-ci.com :+1:

johnynek:
Thanks @decause !

nevillelyh:
Build failed because a serialization change but I can't seem to reproduce this locally. Will keep digging.

```
[info] - should serialize as expected to the correct value (see above for details) *** FAILED ***
[info]   "TgFhKuI[A]" did not equal "TgFhKuI[=]" a*b with serialization id 76 serializes to TgFhKuIA, but the test example is TgFhKuI= (SerializedExamplesOfStandardDataSpec.scala:281)
```

nevillelyh:
Finally got everything green. @johnynek WDYT?

@regadas could use another pair of 👀 when you get a chance.

regadas:
@nevillelyh roger, I'll have a look!

chill-java/src/main/java/com/twitter/chill/java/RegexSerializer.java
```diff
@@ -35,11 +35,10 @@ static public IKryoRegistrar registrar() {
     @Override
     public void write(Kryo kryo, Output output, Pattern pattern) {
         output.writeString(pattern.pattern());
-        output.writeInt(pattern.flags(), true);
     }
 
     @Override
     public Pattern read(Kryo kryo, Input input, Class<Pattern> patternClass) {
-        return Pattern.compile(input.readString(), input.readInt(true));
+        return Pattern.compile(input.readString());
```
regadas, line 42: This seems the revert in #331; should we rebase?

.travis.yml
```diff
@@ -3,21 +3,22 @@ sudo: false
...
-      script: ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
+      jdk: openjdk8
+      script: sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
       after_success:
         - bash <(curl -s https://codecov.io/bash)
 
-    - scala: 2.12.6
-      jdk: oraclejdk8
-      script: ./sbt "+++$TRAVIS_SCALA_VERSION clean" "+++$TRAVIS_SCALA_VERSION test" # "+++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
+    - scala: 2.12.8
```
regadas, line 15: saw that you had issues with `2.12.10` in 2dfc5bd; curious if the same thing happens with `2.12.9`

regadas (approved):
LGTM

.travis.yml
nevillelyh: Yes happens to both 2.12.9 & 2.12.10, have to stick with 2.12.8.

chill-java/src/main/java/com/twitter/chill/java/RegexSerializer.java
nevillelyh: #331 is actually a breaking change since it appends an additional int to the serialized bytes. This is to revert that change.

regadas: ahh ok 😄, i thought that was introduced in #326 and #331 was supposed to reverted it but closed due to build failure. Anyway good that it's reverted! 👍

PR #351, opened 2019-11-05
Clean closures transitively

regadas:
Picked up PR #292.

* Added a new ClosureCleaner
* Added MiMa exclusion rules since new ClosureCleaner has different API.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -151,14 +151,12 @@ val ignoredABIProblems = {
   import com.typesafe.tools.mima.core.ProblemFilters._
   Seq(
     exclude[MissingTypesProblem]("com.twitter.chill.storm.BlizzardKryoFactory"),
-    exclude[IncompatibleResultTypeProblem]("com.twitter.chill.ClosureCleaner.*"),
-    exclude[DirectMissingMethodProblem]("com.twitter.chill.ClosureCleaner.*"),
-    exclude[IncompatibleMethTypeProblem]("com.twitter.chill.ClosureCleaner.*"),
-    exclude[IncompatibleTemplateDefProblem]("com.twitter.chill.ClosureCleaner"),
-    exclude[FinalClassProblem]("com.twitter.chill.InnerClosureFinder"),
-    exclude[DirectMissingMethodProblem]("com.twitter.chill.InnerClosureFinder.*"),
-    exclude[IncompatibleResultTypeProblem]("com.twitter.chill.InnerClosureFinder.*"),
-    exclude[MissingClassProblem]("com.twitter.chill.FieldAccessFinder")
+    exclude[MissingTypesProblem]("com.twitter.chill.InnerClosureFinder"),
+    exclude[IncompatibleResultTypeProblem]("com.twitter.chill.InnerClosureFinder.visitMethod"),
+    exclude[IncompatibleResultTypeProblem]("com.twitter.chill.FieldAccessFinder.visitMethod"),
+    exclude[MissingClassProblem]("com.twitter.chill.FieldAccessFinder"),
+    exclude[MissingTypesProblem]("com.twitter.chill.FieldAccessFinder"),
+    exclude[DirectMissingMethodProblem]("com.twitter.chill.FieldAccessFinder.this")
   )
 }
 
--- a/build.sbt
+++ b/build.sbt
@@ -150,7 +150,15 @@ val ignoredABIProblems = {
   import com.typesafe.tools.mima.core._
   import com.typesafe.tools.mima.core.ProblemFilters._
   Seq(
-    exclude[MissingTypesProblem]("com.twitter.chill.storm.BlizzardKryoFactory")
+    exclude[MissingTypesProblem]("com.twitter.chill.storm.BlizzardKryoFactory"),
+    exclude[IncompatibleResultTypeProblem]("com.twitter.chill.ClosureCleaner.*"),
+    exclude[DirectMissingMethodProblem]("com.twitter.chill.ClosureCleaner.*"),
+    exclude[IncompatibleMethTypeProblem]("com.twitter.chill.ClosureCleaner.*"),
+    exclude[IncompatibleTemplateDefProblem]("com.twitter.chill.ClosureCleaner"),
+    exclude[FinalClassProblem]("com.twitter.chill.InnerClosureFinder"),
+    exclude[DirectMissingMethodProblem]("com.twitter.chill.InnerClosureFinder.*"),
+    exclude[IncompatibleResultTypeProblem]("com.twitter.chill.InnerClosureFinder.*"),
+    exclude[MissingClassProblem]("com.twitter.chill.FieldAccessFinder")
   )
 }
 
@@ -177,6 +185,7 @@ lazy val chill = Project(
   .settings(
     name := "chill",
     mimaPreviousArtifacts := Set("com.twitter" %% "chill" % binaryCompatVersion),
+    mimaBinaryIssueFilters ++= ignoredABIProblems,
     libraryDependencies += "org.apache.xbean" % "xbean-asm7-shaded" % asmVersion
   )
   .dependsOn(chillJava)
--- a/build.sbt
+++ b/build.sbt
@@ -6,6 +6,7 @@ val algebirdVersion = "0.13.5"
 val bijectionVersion = "0.9.6"
 val kryoVersion = "4.0.2"
 val scroogeVersion = "4.12.0"
+val asmVersion = "4.15"
 
 val sharedSettings = mimaDefaultSettings ++ Seq(
   organization := "com.twitter",
@@ -175,7 +176,8 @@ lazy val chill = Project(
 ).settings(sharedSettings)
   .settings(
     name := "chill",
-    mimaPreviousArtifacts := Set("com.twitter" %% "chill" % binaryCompatVersion)
+    mimaPreviousArtifacts := Set("com.twitter" %% "chill" % binaryCompatVersion),
+    libraryDependencies += "org.apache.xbean" % "xbean-asm7-shaded" % asmVersion
   )
   .dependsOn(chillJava)
 
--- a/chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
+++ b/chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
@@ -30,18 +30,15 @@
 package com.twitter.chill
 
 import _root_.java.lang.reflect.Field
+import _root_.java.io.{ByteArrayInputStream, ByteArrayOutputStream, ObjectInputStream, ObjectOutputStream}
 
-import scala.collection.mutable.{Set => MSet, Map => MMap}
+import org.apache.xbean.asm7.Opcodes._
+import org.apache.xbean.asm7.{ClassReader, ClassVisitor, MethodVisitor, Type}
 
 import scala.annotation.tailrec
-
-import com.esotericsoftware.reflectasm.shaded.org.objectweb.asm.{
-  ClassReader,
-  ClassVisitor,
-  MethodVisitor,
-  Type
-}
-import com.esotericsoftware.reflectasm.shaded.org.objectweb.asm.Opcodes._
+import scala.collection.mutable.{Map => MMap, Set => MSet, Stack => MStack}
+import scala.language.existentials
+import scala.util.Try
 
 /**
  * Copied from Spark, written by Matei Zaharia (matei@cs.berkeley.edu).
@@ -60,15 +57,35 @@ object ClosureCleaner {
   private val innerClasses = MMap[Class[_], Set[Class[_]]]()
   private val accessedFieldsMap = MMap[Class[_], Set[Field]]()
 
-  private def getClassReader(cls: Class[_]): ClassReader = {
-    val className = cls.getName.replaceFirst("^.*\\.", "") + ".class"
-    new ClassReader(cls.getResourceAsStream(className))
+  private[chill] def serialize[T](t: T): Array[Byte] = {
+    val bos = new ByteArrayOutputStream
+    val out = new ObjectOutputStream(bos)
+    try {
+      out.writeObject(t)
+      bos.toByteArray
+    } finally {
+      out.close()
+      bos.close()
+    }
+  }
+
+  private[chill] def deserialize[T](bytes: Array[Byte]): T = {
+    val bis = new ByteArrayInputStream(bytes)
+    val in = new ObjectInputStream(bis);
+    try {
+      in.readObject.asInstanceOf[T]
+    } finally {
+      bis.close()
+      in.close()
+    }
   }
 
   // Return the $outer field for this class
   def outerFieldOf(c: Class[_]): Option[Field] =
     outerFields
-      .getOrElseUpdate(c, c.getDeclaredFields.find { _.getName == OUTER })
+      .getOrElseUpdate(c, c.getDeclaredFields.find {
+        _.getName == OUTER
+      })
 
   /**
    * this does reflection each time
@@ -78,10 +95,9 @@ object ClosureCleaner {
   private def getOuterClassesFn(cls: Class[_], hierarchy: List[Class[_]] = Nil): List[Class[_]] =
     outerFieldOf(cls) match {
       case None => hierarchy
-      case Some(f) => {
+      case Some(f) =>
         val next = f.getType
         getOuterClassesFn(next, next :: hierarchy)
-      }
     }
 
   def outerClassesOf(cls: Class[_]): List[Class[_]] =
@@ -94,8 +110,8 @@ object ClosureCleaner {
   @tailrec
   def getOutersOf(obj: AnyRef, hierarchy: List[(Class[_], AnyRef)] = Nil): List[(Class[_], AnyRef)] =
     outerFieldOf(obj.getClass) match {
-      case None => hierarchy // We have finished
-      case Some(f) => {
+      case None    => hierarchy // We have finished
+      case Some(f) =>
         // f is the $outer of obj
         f.setAccessible(true)
         // myOuter = obj.$outer
@@ -104,20 +120,20 @@ object ClosureCleaner {
         // Note that if you use f.getType you might get an interface. No good
         val outerType = myOuter.getClass
         getOutersOf(myOuter, (outerType, myOuter) :: hierarchy)
-      }
     }
 
   private def getInnerClassesFn(inCls: Class[_]): Set[Class[_]] = {
     val seen = MSet[Class[_]](inCls)
-    var stack = List[Class[_]](inCls)
-    while (!stack.isEmpty) {
-      val cr = getClassReader(stack.head)
-      stack = stack.tail
+    val stack = MStack[Class[_]](inCls)
+    while (stack.nonEmpty) {
+      // mutated by InnerClosureFinder
       val set = MSet[Class[_]]()
-      cr.accept(new InnerClosureFinder(set), 0)
-      for (cls <- set -- seen) {
+      AsmUtil.classReader(stack.pop()).foreach { cr =>
+        cr.accept(new InnerClosureFinder(set), 0)
+      }
+      (set -- seen).foreach { cls =>
         seen += cls
-        stack = cls :: stack
+        stack.push(cls)
       }
     }
     (seen - inCls).toSet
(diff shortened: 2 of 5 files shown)


chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
```diff
@@ -29,221 +29,238 @@
...
-  private def getClassReader(cls: Class[_]): ClassReader = {
-    val className = cls.getName.replaceFirst("^.*\\.", "") + ".class"
-    new ClassReader(cls.getResourceAsStream(className))
+  /** Clean the closure in place. */
+  final def apply[T <: AnyRef](func: T): T = {
+    try {
+      val buffer = new ByteArrayOutputStream
+      val oos = new ObjectOutputStream(buffer)
+      try {
+        oos.writeObject(func.asInstanceOf[Serializable])
```
johnynek: do you want to also deserialize it? I have seen cases where write succeeds, but read somehow fails.

chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
```diff
@@ -29,221 +29,238 @@
...
+      } finally {
+        if (oos != null) oos.close()
+      }
+      buffer.toByteArray
```
johnynek: why do this?

chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
```diff
@@ -29,221 +29,238 @@
...
-    new ClassReader(cls.getResourceAsStream(className))
+  /** Clean the closure in place. */
+  final def apply[T <: AnyRef](func: T): T = {
+    try {
+      val buffer = new ByteArrayOutputStream
+      val oos = new ObjectOutputStream(buffer)
+      try {
+        oos.writeObject(func.asInstanceOf[Serializable])
+      } finally {
+        if (oos != null) oos.close()
```
johnynek: how can this be null? Line 52 assigns it to a non-null value?

chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
```diff
@@ -29,221 +29,238 @@
...
-        // Add all of af:
-        af.foreach { clsMSet =>
-          val set = clsMSet._2.toSet
-          accessedFieldsMap += ((clsMSet._1, toF(set)))
+  def innerClassesOf(func: AnyRef): Set[Class[_]] = {
+    val seen = MSet[Class[_]](func.getClass)
+    val stack = MStack[Class[_]](func.getClass)
+    while (stack.nonEmpty) {
+      val cr = AsmUtil.classReader(stack.pop())
+      val set = MSet[Class[_]]()
```
johnynek, line 130: can we comment that this is mutated by InnerClosureFinder. If you forget that, like 104 can never be entered. That said, we don't exercise entering it from the looks of the code coverage.

chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
```diff
@@ -29,221 +29,238 @@
...
+}
+
+sealed trait ClosureCleaner {
+  import ClosureCleaner._
+
+  /** The closure to clean. */
+  val func: AnyRef
+
+  /** Clean [[func]] by replacing its 'outer' with a cleaned clone. See [[cleanOuter()]]. */
+  def clean: AnyRef = {
```
johnynek: why not `def clean(): Unit` since this has a side effect? and make it clearer that you should read `func` after?

chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
```diff
@@ -29,221 +29,238 @@
...
+  }
+
+  /**
+   * Create a new 'cleaned' copy of [[func]]'s outer, without modifying the original. The cleaned
+   * outer may have null values for fields that are determined to be unneeded in the context
+   * of [[func]].
+   *
+   * @return The cleaned outer.
+   */
+  def cleanOuter(): AnyRef
```
johnynek: should this be `def cleanOuter: AnyRef` since it is a pure function?

chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
```diff
@@ -29,221 +29,238 @@
...
+ * An implementation of [[ClosureCleaner]] that cleans [[func]] by transitively tracing its method
+ * calls and field references up through its enclosing scopes. A new hierarchy of outers is
+ * constructed by cloning the outer scopes and populating only the fields that are to be accessed
+ * by [[func]] (including any of its inner closures). Additionally, outers are removed from the
+ * new hierarchy if none of their fields are accessed by [[func]].
+ */
+private final class TransitiveClosureCleaner(val func: AnyRef) extends ClosureCleaner {
+  import ClosureCleaner._
+
+  private lazy val outerClasses: List[(Class[_], AnyRef)] = outerClassesOf(func)
```
johnynek: can you comment why lazy? Seems we will always evaluate these in normal operation and no reason they can't be just `private val`?

chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
```diff
@@ -29,221 +29,238 @@
...
+          }
 
-      override def visitMethodInsn(op: Int, owner: String, name: String, desc: String, itf: Boolean) {
-        // Check for calls a getter method for a variable in an interpreter wrapper object.
-        // This means that the corresponding field will be accessed, so we should save it.
-        if (op == INVOKEVIRTUAL && owner.endsWith("$iwC") && !name.endsWith("$outer"))
-          for (cl <- output.keys if cl.getName == owner.replace('/', '.'))
-            output(cl) += name
+        override def visitMethodInsn(op: Int, owner: String, name: String, desc: String, itf: Boolean): Unit =
+          for (cl <- output.keys if cl.getName == owner.replace('/', '.')) {
```
johnynek: I think this `replace` runs on each call. Can we make it a val:
```
val ownerName = owner.replace('/', '.')
output
  .keys
  .iterator
  .filter(_.getName == ownerName)
  .foreach { cl =>

  }
```
would actually be my preference...

chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
```diff
@@ -29,221 +29,238 @@
...
+              }
+            }
+          }
       }
     }
 }
 
-class InnerClosureFinder(output: MSet[Class[_]]) extends ClassVisitor(ASM5) {
-  var myName: String = null
+private final class InnerClosureFinder(output: MSet[Class[_]]) extends ClassVisitor(ASM7) {
```
johnynek: can you add a comment as to what this is supposed to be doing?

chill-scala/src/test/scala/com/twitter/chill/BaseProperties.scala
```diff
@@ -32,12 +33,14 @@ trait BaseProperties {
 
   def rtEquiv[T](t: T): Boolean = {
     val serdeser = rt(t)
-    serdeser == t && serdeser.getClass.asInstanceOf[Class[Any]] == t.getClass.asInstanceOf[Class[Any]]
+    serdeser == t && serdeser.getClass.asInstanceOf[Class[Any]] == t.getClass
```
johnynek: can we add a parens, this reads weird: `t.getClass.asInstanceOf[Class[Any]]` also, I'm not sure why we need the cast to use `==` Can we remove both casts?

chill-scala/src/test/scala/com/twitter/chill/BaseProperties.scala
```diff
@@ -32,12 +33,14 @@ trait BaseProperties {
...
+      .asInstanceOf[Class[Any]]
   }
 
   def rtEquiv[T](k: KryoInstantiator, t: T): Boolean = {
     val serdeser = rt(k, t)
-    serdeser == t && serdeser.getClass.asInstanceOf[Class[Any]] == t.getClass.asInstanceOf[Class[Any]]
+    serdeser == t && serdeser.getClass.asInstanceOf[Class[Any]] == t.getClass
```
johnynek: same comment as above.

chill-scala/src/test/scala/com/twitter/chill/BaseProperties.scala
```diff
@@ -48,19 +51,20 @@ trait BaseProperties {
...
+      out.close()
+      bos.close()
     }
   }
   def jdeserialize[T](bytes: Array[Byte])(implicit cmf: ClassTag[T]): T = {
     val cls = cmf.runtimeClass.asInstanceOf[Class[T]]
     val bis = new ByteArrayInputStream(bytes)
     val in = new ObjectInputStream(bis);
     try {
+      println(cls.getName)
```
johnynek: can we remove the println?

chill-scala/src/test/scala/com/twitter/chill/ClosureCleanerSpec.scala
```diff
@@ -70,5 +75,33 @@ class ClosureCleanerSpec extends WordSpec with Matchers {
       ClosureCleaner(fn)
       fn(10) should equal(420)
     }
+
+    "Handle nested closures in non-serializable object" in {
+      val fn: Int => Int = new NestedClosuresNotSerializable().getMapFn
+
+      ClosureCleaner(fn)
+      isSerializable(fn) shouldBe true
+      fn(6) shouldEqual 8
```
johnynek: can we use scalacheck here to do:
```
val fn0 = new NestedClosuresNotSerializable().getMapFn
val fn1 = new NestedClosuresNotSerializable().getMapFn
ClosureCleaner(fn1)
forAll { i: Int =>
 assert(fn1(i) == fn0(i))
}
// this wouldn't serialize without cleaning.
assert(!isSerializable(fn0))
```

johnynek (commented):
Thanks for the PR here. It is pretty subtle code.

Take a look here:
https://codecov.io/gh/twitter/chill/pull/351/diff?src=pr&el=tree#diff-Y2hpbGwtc2NhbGEvc3JjL21haW4vc2NhbGEvY29tL3R3aXR0ZXIvY2hpbGwvQ2xvc3VyZUNsZWFuZXIuc2NhbGE=

The coverage isn't very high. There are a lot of blocks of the code not covered by the tests. Do you think we could add more tests to get the coverage up so we can see all the branches work?

The law I kind of want is something like this:
```
def law[A: Arbitrary, B](fn: => A => B) = {
  val clean = ClosureCleaner(fn)
  val fn0 = fn
  assert(!isSerializable(fn0))
  assert(isSerializable(clean))
  forAll { a: A =>
    assert(clean(a) == fn0(a))
  }
} 
```
Then just add several calls:
```
law(new SomeFn())
law(new SomeOtherFn)
...
```

build.sbt
```diff
@@ -142,7 +143,15 @@ val ignoredABIProblems = {
   import com.typesafe.tools.mima.core._
   import com.typesafe.tools.mima.core.ProblemFilters._
   Seq(
-    exclude[MissingTypesProblem]("com.twitter.chill.storm.BlizzardKryoFactory")
+    exclude[MissingTypesProblem]("com.twitter.chill.storm.BlizzardKryoFactory"),
+    exclude[IncompatibleResultTypeProblem]("com.twitter.chill.ClosureCleaner.*"),
```
johnynek: how inconvenient do you think it would be to make this binary compatible? Binary incompatibility is a pain if we could use `abstract class` to fix?

chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
regadas: yeah it's better to make sure 👍

chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala
regadas: smth went terribly wrong here ... 🤔

regadas:
Hi @johnynek! Thanks for the review and sorry for late reply!

I thought about binary compatibility and decided to not add a new `ClosureCleaner` but fix the existing one instead. I still added a some `MiMa` exclusion rules due to moving from `ASM5` to `ASM7`.

I also reworked `ClosureCleanerSpec` with the addition of serialization laws and takes into account scala `2.11` /  `2.12` lambda implementations. Code coverage is actually different between the two being a little bit higher with `2.11` since there are some cases where `ClosureCleaner` is not needed with `2.12` (current report is for 2.12).

chill-scala/src/test/scala/com/twitter/chill/BaseProperties.scala
regadas: I can have a look

regadas:
`rebase` with `develop`

regadas:
CI seems stuck...

johnynek:
I'd really love to find a way to get 100% test coverage on this. Actually, scalding doesn't use this closure cleaner. It didn't seem to help in many of the cases (maybe we had bugs?) that caused problems. In most cases we saw, it was like this:

```
object Foo {
  val bar = ..
  // this is not actually used by any closure and is not serializable
  val boom = ... 

  // we really need outer because we access this.bar
  val fn = { x: Quux => f(bar, x) }
```
So, the closure cleaner, I think, isn't powerful enough to solve this case. Is it? If it is, it would be good to have a test that exercises and describes this case.

I really appreciate your contributions here, and I hate to make better the enemy of good, so if you really don't have the time to add test coverage, I think things are at least better, but it would be really nice to use the coverage information to try to go for 100% and exercise all the code paths or comment why we think they are unreachable in practice.

chill-scala/src/test/scala/com/twitter/chill/ClosureCleanerSpec.scala
```diff
@@ -16,59 +16,227 @@ limitations under the License.
...
-    "Handle captured vals" in {
-      val answer = 42
-      val fn = { x: Int =>
-        answer * x
-      }
-      ClosureCleaner(fn)
-      fn(10) should equal(420)
+
+    "clean outers with objects" in {
+      serializableFn(TestObject.fn, before = true, after = true)
```
johnynek, line 157: how do we know this is doing anything if `before = true`? Can we find a case with an object where `before = false`?

chill-scala/src/test/scala/com/twitter/chill/ClosureCleanerSpec.scala
```diff
@@ -16,59 +16,227 @@ limitations under the License.
...
       }
 
       val t = new Test("you all everybody")
-      val fn = t.rev _
-      //debug(fn)
-      //println(ClosureCleaner.getOutersOf(fn))
-      ClosureCleaner(fn)
-      fn("hey") should equal(20)
+      val fn: String => Int = t.rev
+      serializableFn(fn, before = false, after = false)
```
johnynek, line 146: there is no use of ClosureCleaner in this test. Does it not work with this case?

chill-scala/src/test/scala/com/twitter/chill/ClosureCleanerSpec.scala
```diff
@@ -16,59 +16,227 @@ limitations under the License.
...
+            Option(y).foreach { z =>
+              someSerializableValue
+            }
+          }
+        }
+      }
+
+      serializableFn(closure1, before = false, after = false)
+      serializableFn(closure2, before = false, after = false)
+      serializableFn(closure3, before = false, after = false)
```
johnynek, line 135: we can't support these cases? What is the goal of this test? Why can't these cases be supported, can you add a comment?

chill-scala/src/test/scala/com/twitter/chill/ClosureCleanerSpec.scala
```diff
@@ -16,59 +16,227 @@ limitations under the License.
...
+      val localVal = 1
+      val closure1 = (_: Int) => localVal
+      val closure2 = (x: Int) => x + localVal
+
+      serializableFn(closure1, before = true, after = true)
+      serializableFn(closure2, before = true, after = true)
+    }
+
+    "clean basic closures (non LMF)" in {
+      assume(!ClosureCleanerSpec.supportsSerializedLambda)
```
johnynek, line 57: I don't know what this assume construct does. I think it is the first use in these tests. Is it bailing out of the test if that is not true?

regadas: due to the different closures implementations between 2.11 (anonymous class) and 2.12 (java.lang.invoke.LambdaMetaFactory) use cases in 2.11 that require closure cleaning in 2.12 is not needed. With the usage of `assume` we can actually "CANCEL" (avoids) the test execution if the predicate is true.

chill-scala/src/test/scala/com/twitter/chill/ClosureCleanerSpec.scala
regadas: This is an interesting test ... we have 2 tests with the same thing; closures reference properties in ClosureCleanerSpec; one test is non-LMF and ther other LMF. However under 2.11 (non-LMF) we have access to ther $outer ref and we can clean it but with 2.12 LMF Lambda has no outer ref; I couldn't find a way to clean these cases yet.

chill-scala/src/test/scala/com/twitter/chill/ClosureCleanerSpec.scala
regadas: ClosureCleaner is applied in serializableFn (maybe we should change the name?) yeah this is a strange case i can't manage to find an $outer reference in either 2.11 and 2.12

chill-scala/src/test/scala/com/twitter/chill/ClosureCleanerSpec.scala
regadas: yeah good point; in this case I guess the thought that if something is serializable before after applying `ClosureCleaner` it should still be serializable. I tried to find a case where it fails but I couldn't come up with one.

johnynek:
@decause can you help with the CLA bot here?

Note in #292 @andrewsmartin signed the CLA, and @regadas has also signed.

I'd like to merge this PR, but I can't do so due the CLA bot issues.

chill-scala/src/test/scala/com/twitter/chill/ClosureCleanerSpec.scala
johnynek: okay, in any case, we seem to have made progress.

regadas: Yeah we made some progress here. As soon as we release this I can actually use it in https://github.com/spotify/scio

regadas:
sorry about this, rebasing with develop due to the scalafix changes went a bit sideways :/

johnynek:
Looks like we still have a merge conflict.

decause:
> [CLA assistant check](https://cla-assistant.io/twitter/chill?pullRequest=351)
> Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/twitter/chill?pullRequest=351) before we can accept your contribution.
> **1** out of **2** committers have signed the CLA.
> 
> white_check_mark regadas
> x andrewsmartin
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/twitter/chill?pullRequest=351) it.

I've opened an issue upstream to look into this, but tl;dr, I can find @andrewsmartin as having signed the CLA for the twitterdev org, but not our primary twitter org. Since I have manually confirmed that @andrewsmartin has signed the CLA, this PR can be merged in the meantime if it is blocking your team, while we work with CLA-Assistant to confirm the checking behavior/migration issue/re-signing resolution.

Thank you all for your continued support <3

johnynek:
Thanks for the work on this everyone!

regadas:
Thank you!

regadas:
@johnynek when do you think it's possible to have a new chill release?

johnynek:
I can try to publish on Friday.... let's see if I can brave the sbt publishing daemons or if I will once again swear off publishing code...

I know there have been some new releases to sbt-sonatype... and there is sbt-release... somehow I need to install and keep up to date several bespoke plugins and keep abreast of their changes.... but maybe I will be successful in publishing.

nevillelyh:
I talked to @regadas IRL, the least we can do is to publish to Sonatype staging under our org, just to verify that all sbt stuff are correct. You might still have to sort out GPG/PGP/pinentry on your end but I just went through that myself so 🤞

regadas:
cool! feel free to poke as well if you need anything!

PR #361, opened 2019-11-25
Update scala to 2.12.10

regadas:
The issue around specialized tuples and newer scala versions was raised [here](https://github.com/twitter/chill/pull/330#pullrequestreview-288959050).

This PR doesn't keep binary compatibility due to type signature changes, however the generated classes are almost the same™ when disassembled. This might not be that big of a deal since the users will most likely not use them directly.

@johnynek wdyt?


johnynek:
won't this be a serious performance regression? If we merge this the types that were specialized will not have to add tags for each element in the tuple (doubling in many cases the size of the data).

I think such a change is probably an unacceptable loss just to upgrade a minor version of the compiler.

If the classes still exist, maybe we need to generate java rather than scala code, but I think we should be very circumspect about giving up this win.

regadas:
Hi @johnynek so with this we still keep the specialized mapping: 

```scala
newK.register(Class.forName("scala.Tuple1$mcJ$sp"), new Tuple1LongSerializer)
```

```scala
class Tuple1LongSerializer extends KSerializer[Tuple1[Long]] with Serializable {
   setImmutable(true)
   def read(kser: Kryo, in: Input, cls: Class[Tuple1[Long]]): Tuple1[Long] =
     new Tuple1[Long](in.readLong)
   def write(kser: Kryo, out: Output, tup: Tuple1[Long]): Unit =
     out.writeLong(tup._1)
 }
```

this class when disassembled looks the same the exception of the extends signature... Looking the same means that it doing `tup._1` is in reality using the specialized method. 

```
  public void write(com.esotericsoftware.kryo.Kryo, com.esotericsoftware.kryo.io.Output, scala.Tuple1<java.lang.Object>);
    descriptor: (Lcom/esotericsoftware/kryo/Kryo;Lcom/esotericsoftware/kryo/io/Output;Lscala/Tuple1;)V
    Code:
       0: aload_2
       1: aload_3
       2: invokevirtual #40                 // Method scala/Tuple1._1$mcJ$sp:()J
       5: invokevirtual #45                 // Method com/esotericsoftware/kryo/io/Output.writeLong:(J)V
       8: return
```

regadas:
@johnynek it's totally possible that I'm missing something or miss-understanding 😄  but this solution seems fairly equivalent. The idea is to take this approach and use it for 2.13.1 which I intend to start soon.

build.sbt
```diff
@@ -157,7 +178,11 @@ val ignoredABIProblems = {
     exclude[IncompatibleResultTypeProblem]("com.twitter.chill.FieldAccessFinder.visitMethod"),
     exclude[MissingClassProblem]("com.twitter.chill.FieldAccessFinder"),
     exclude[MissingTypesProblem]("com.twitter.chill.FieldAccessFinder"),
-    exclude[DirectMissingMethodProblem]("com.twitter.chill.FieldAccessFinder.this")
+    exclude[DirectMissingMethodProblem]("com.twitter.chill.FieldAccessFinder.this"),
+    exclude[IncompatibleResultTypeProblem]("com.twitter.chill.Tuple1*Serializer.read"),
+    exclude[IncompatibleMethTypeProblem]("com.twitter.chill.Tuple1*Serializer.write"),
+    exclude[IncompatibleResultTypeProblem]("com.twitter.chill.Tuple2*Serializer.read"),
+    exclude[IncompatibleMethTypeProblem]("com.twitter.chill.Tuple2*Serializer.write")
```
johnynek, line 185: these require a version bump to `0.10.0` before we publish next. Can we not break compatibility?

johnynek (commented):
okay, i misunderstood the fact that the scala compiler was translating these types into the specialized types so that the jvm bytecode does not change.

That said, can we avoid the mima incompatibility? That would be ideal.

build.sbt
regadas: yeah that's the sad part ... I couldn't find a way to not break it unfortunately. I know it's kinda of annoying to have this version bump due to a minor scala version update; the bright side is that this should give us better compatibility for future versions since we don't reference internal classes directly.

regadas: @johnynek do you think it's better to bump version in a separate PR?

johnynek:
Sorry for the delay.

I’m really afraid this is going to burn people with runtime exceptions on spark and scalding (and scio) but you all are putting in the work so I guess we’ll go forward.

regadas:
@johnynek no need to apologize, time is short! 😄 Thanks!

regadas:
I'll start working on scala 2.13 support soon™

PR #362, opened 2019-11-25
Fix ScalaKryoInstantiator return type

regadas:
This kinda related to #359. ~Separated it because this one requires excluding MiMa and it's open for discussion.~ scratch that there's no issue at all.


johnynek:
:+1:

johnynek (approved):

PR #370, opened 2019-12-05
Update algebird to 0.13.6

regadas:


regadas:
ahhh oops ... forgot about 2.10. I'll bump algebird as part of 2.13 update. Should we also drop 2.10 with that update?

johnynek:
Yeah, I think 2.10 is very long in the tooth.

regadas:
ack 👍

PR #372, opened 2019-12-05
Add scala 2.13 support

regadas:


regadas:
Opening this up ... I might have missed smth. @johnynek @nevillelyh can you take a look?

chill-scala/src/main/scala-2.12-/com/twitter/chill/ScalaCollectionsRegistrarCompat.scala
```diff
@@ -0,0 +1,21 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+
+package com.twitter.chill
+
+class ScalaCollectionsRegistrarCompat extends IKryoRegistrar {
```
johnynek, line 26: why does this need to be in 2.12-? looks like it doesn't use any 2.12isms?

chill-scala/src/main/scala-2.13+/com/twitter/chill/ScalaCollectionsRegistrarCompat.scala
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+ */
+
+package com.twitter.chill
+
+import scala.collection.immutable.Range
+
+class ScalaCollectionsRegistrarCompat extends IKryoRegistrar {
+  override def apply(newK: Kryo): Unit =
+    newK.register(classOf[Range.Exclusive])
```
johnynek, line 30: can we comment the difference here? I didn't notice why we would separate before.

chill-scala/src/main/scala-2.12-/com/twitter/chill/ScalaCollectionsRegistrarCompat.scala
johnynek: feels weird that this doesn't actually do anything.

johnynek (approved):
looks good to me, just a few comments.

chill-scala/src/main/scala-2.12-/com/twitter/chill/ScalaCollectionsRegistrarCompat.scala
regadas: Yeah. This is here to avoid doing scala version checks to add the
`newK.register(classOf[Range.Exclusive])` for `2.13` Maybe we should add this as comment?

chill-scala/src/main/scala-2.13+/com/twitter/chill/ScalaCollectionsRegistrarCompat.scala
regadas: In `2.12` `Range` is exclusive by default but it doesn't return a specific type.
```scala
scala> Range(1, 3)
res4: scala.collection.immutable.Range = Range 1 until 3

scala> Range.inclusive(1, 3)
res5: scala.collection.immutable.Range.Inclusive = Range 1 to 3
```

In `2.13`
```scala
scala> Range(1, 3)
res0: scala.collection.immutable.Range.Exclusive = Range 1 until 3

scala> Range.inclusive(1, 3)
res1: scala.collection.immutable.Range.Inclusive = Range 1 to 3
```

chill-scala/src/main/scala-2.12-/com/twitter/chill/ScalaCollectionsRegistrarCompat.scala
johnynek: Yeah I think explaining this for the person working on 2.14 who is confused it would be useful.

johnynek:
Thanks!

Issue #421, opened 2020-04-22
Java8 lambda expression in PriorityQueue cannot be serialized


Jiayi-Liao:
When I'm using `PriorityQueueSerializer` from chill-java package, I got NPE when serializing `PriorityQueue` instance like this:

```
PriorityQueue<String> pq = new PriorityQueue<>((o1, o2) -> o1.length() - o2.length() - 1);
pq.add("1234135");
pq.add("12323424135");
```

```
java.lang.NullPointerException
	at com.esotericsoftware.kryo.util.DefaultClassResolver.writeClass(DefaultClassResolver.java:80)
	at com.esotericsoftware.kryo.Kryo.writeClass(Kryo.java:488)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:593)
```
Perhaps it's because the lambda expression is not serializable. But the strange thing is that when I removed the registration of `PriorityQueueSerializer` from kryo's registrations, the serialization works.

Is this expected?

johnynek:
thanks for the bug report. This isn't expected. So, this is a bug. Would you be willing to open a PR with a test that shows the bug?

johnynek:
note, we have a test here:

https://github.com/twitter/chill/blob/develop/chill-java/src/test/scala/com/twitter/chill/java/PriorityQueueTest.scala

johnynek:
here is the serializer: https://github.com/twitter/chill/blob/develop/chill-java/src/main/java/com/twitter/chill/java/PriorityQueueSerializer.java

Jiayi-Liao:
@johnynek Hi, I've created a PR to reproduce the problem (not a NPE but the same strange phenomenon). Could you take a look ?

By the way I notice that this project is still using java 1.6, does it mean that the project doesn't support java 1.8 and also dosen't take the java 1.8 lambda expression into consideration?

PR #423, opened 2020-04-27
Reproduce exception with PriorityQueue with lambda expression

Jiayi-Liao:
The pull request is to reproduce the exception described in https://github.com/twitter/chill/issues/421

Set the java version to 1.8 because the problem is about java 1.8 lambda expression.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -31,8 +31,8 @@ val sharedSettings = mimaDefaultSettings ++ Seq(
   scalariformPreferences := formattingPreferences,
   scalariformAutoformat := false,
   // Twitter Hadoop needs this, sorry 1.7 fans
-  javacOptions ++= Seq("-target", "1.6", "-source", "1.6", "-Xlint:-options"),
-  javacOptions in doc := Seq("-source", "1.6"),
+  javacOptions ++= Seq("-target", "1.8", "-source", "1.8", "-Xlint:-options"),
+  javacOptions in doc := Seq("-source", "1.8"),
   resolvers ++= Seq(
     Opts.resolver.sonatypeSnapshots,
     Opts.resolver.sonatypeReleases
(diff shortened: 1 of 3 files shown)


nevillelyh:
Some builds failed because of sbt classloading:

```
[error] Test suite com.twitter.chill.java.PriorityQueueSpec failed with java.lang.ClassNotFoundException: com.twitter.chill.java.TestPriorityQueue$$Lambda$6275/1415420090.
[error] This may be due to the ClassLoaderLayeringStrategy (ScalaLibrary) used by your task.
[error] To improve performance and reduce memory, sbt attempts to cache the class loaders used to load the project dependencies.
[error] The project class files are loaded in a separate class loader that is created for each test run.
[error] The test class loader accesses the project dependency classes using the cached project dependency classloader.
[error] With this approach, class loading may fail under the following conditions:
[error] 
[error]  * Dependencies use reflection to access classes in your project's classpath.
[error]    Java serialization/deserialization may cause this.
[error]  * An open package is accessed across layers. If the project's classes access or extend
[error]    jvm package private classes defined in a project dependency, it may cause an IllegalAccessError
[error]    because the jvm enforces package private at the classloader level.
[error] 
[error] These issues, along with others that were not enumerated above, may be resolved by changing the class loader layering strategy.
[error] The Flat and ScalaLibrary strategies bundle the full project classpath in the same class loader.
[error] To use one of these strategies, set the  ClassLoaderLayeringStrategy key
[error] in your configuration, for example:
[error] 
[error] set chill-java / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaLibrary
[error] set chill-java / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
[error] 
[error] See ClassLoaderLayeringStrategy.scala for the full list of options.
```

Try adding the suggested settings above?

PR #453, opened 2020-07-12
Update scala versions to 2.12.11 and 2.13.3

regadas:


regadas (commented):
I'm going to split this into separate PR's since 2.13.3 introduces new Vector impl and more underlying changes that change the serialization.

PR #456, opened 2020-07-14
Update to scala 2.13.6

regadas:


chill-scala/src/test/scala-2.13+/com/twitter/chill/SerializedExamplesData.scala
```diff
@@ -148,59 +148,61 @@ object SerializedExamplesData {
...
+    136 -> ("igEBAUgBgmE=" -> ListSet[Any]('a)),
+    137 -> ("iwEBAA==" -> Map[Any, Any]()),
+    138 -> ("jAEBAA==" -> ListMap[Any, Any]()),
+    139 -> ("jQEBAScBSAGCYUgE" -> ListMap('a -> 'a)),
+    140 -> ("jgEBegEBAgI=" -> Stream(1)),
+    141 -> ("jwEB" -> Stream()),
+    142 -> ("kAEBCg==" -> new VolatileByteRef(10)),
+    143 -> ("kQEBAQBqYXZhLm1hdGguQmlnRGVjaW1h7AECAgA=" -> math.BigDecimal(2)),
+    144 -> ("kgEBAA==" -> (Queue.empty[Any], true)),
+    145 -> ("kwEBAQIC" -> (Map(1 -> 2).keySet, true))
```
johnynek: I'm not sure the benefit of these tests if we just change them each time. I think this breaks anyone with stored serialized data. I think storing kryo data long term is a really *REALLY* bad idea, but I think we should have a consistent story here (and maybe bump a major version or something if we really change to "all bets are off!").

regadas: Hi @johnynek, yeah I was thinking the same. After applying this changes I realised that we probably should look into how we actually register classes in kryo since it affects serialization (Kryo registration ids are kept). Maybe we can come up with way to at least incrementally update without breaking. However, If we do need to break it then yeah coming up a new major version should be the way to go.

I'm with you as well, storing kryo data should be avoided. The only use case where i think it's ok is taking snapshots for recovery, while making sure versions stay the same.

regadas: Not sure if we wanna merge this now and tackle what i mentioned afterwards ...

regadas: > I realised that we probably should look into how we actually register classes in kryo since it affects serialization (Kryo registration ids are kept). 

~on a second thought I'll need to spend more time here ... because it might not be as I'm thinking ... in this PR we see that some lower Id's got affected while 94 and 95 were added.~ Just did a quick test and if classes are register incrementally previous ones are not affected.

regadas: > Not sure if we wanna merge this now and tackle what i mentioned afterwards ...

let's wait ... let me see if i come up with smth

chill-scala/src/main/scala-2.13+/com/twitter/chill/AllScalaRegistrarCompat.scala
```diff
@@ -25,7 +25,15 @@ import scala.collection.immutable.Range
...
  * provided registrations.
  */
-class ScalaCollectionsRegistrarCompat extends IKryoRegistrar {
+private[chill] class AllScalaRegistrarCompat_0_9_5 extends IKryoRegistrar {
   override def apply(newK: Kryo): Unit =
     newK.register(classOf[Range.Exclusive])
+
+}
+
+private[chill] class AllScalaRegistrarCompat extends IKryoRegistrar {
```
regadas, line 34: narrowed down the access here ... users shouldn't be using these

chill-scala/src/test/scala-2.13+/com/twitter/chill/SerializedExamplesData.scala
```diff
@@ -197,10 +197,12 @@ object SerializedExamplesData {
     140 -> ("jgEBCg==" -> new VolatileByteRef(10)),
     141 -> ("jwEBAQBqYXZhLm1hdGguQmlnRGVjaW1h7AECAgA=" -> math.BigDecimal(2)),
     142 -> ("kAEBAA==" -> (Queue.empty[Any], true)),
-    143 -> ("kQEBAQIC" -> (Map(1 -> 2).keySet, true))
+    143 -> ("kQEBAQIC" -> (Map(1 -> 2).keySet, true)),
+    144 -> ("kgEBAA==" -> Vector.empty[Int]),
+    145 -> ("kwEBAQIC" -> Vector(1))
```
regadas, line 203: @johnynek now this test data only reflects the new ones

chill-scala/src/test/scala-2.13+/com/twitter/chill/SerializedExamplesData.scala
regadas: @johnynek should be good now!

chill-scala/src/main/scala/com/twitter/chill/ScalaKryoInstantiator.scala
```diff
@@ -207,26 +207,11 @@ class AllScalaRegistrar_0_9_2 extends IKryoRegistrar {
   }
 }
 
-/** Registrar for everything that was registered in chill 0.10.0 */
-class AllScalaRegistrar_0_10_0 extends IKryoRegistrar {
```
regadas, line 211: This class was badly added by me on scala 2.13 migration ... should we keep it around and mark it as deprecated? or can we go ahead and just remove it since we don't expect users to be using it?

regadas:
@johnynek any extra thoughts on this?

johnynek:
one super frustrating thing about this repo is that we try to tell people that they shouldn't use kryo (or chill) for long term serialization (although we don't seem clear about that in the readme, so I guess how would people know that), yet almost every change can break them and we have weak tests for compatibility.

I really wish we could just pick an approach: keep compatibility, or say we are never for long term serialization and each release can break binary serializations.

It makes reviewing pretty tough and anxiety provoking.

See #467 for a recent concern from a user.

regadas:
Hi @johnynek! sorry for late reply! Keeping binary serialization compat will require a lot of effort and a decent test stack as you mentioned. 

I'm leaning towards 2 approaches 

1. We state that we don't guarantee on any release.

2. We do best effort approach where:

* we don't guarantee between scala versions
* we try to keep it between patches and minors but we don't guarantee on majors.
* we don't guarantee on scala version updates (as we know things can change between patches). Maybe here we can go with, on each version update we do a major release.
* we state that long term storage is not recommended and the above.

wdyt?

johnynek:
related to #514 

I think there is a case to be made to drop any guarantee that serialized data is compatible between releases.

What do you think?

regadas:
Hi @johnynek sorry for late reply. Yea it seems there is. I'm gonna update this PR and add notice to the README.

edit: I also share your concerns and I agree with you.

README.md
```diff
@@ -9,6 +9,9 @@ Extensions for the [Kryo serialization library](https://github.com/EsotericSoftw
 serializers and a set of classes to ease configuration of Kryo in systems like Hadoop, Storm,
 Akka, etc.
 
+### Compatibility
```
regadas, line 12: @johnynek added this section. what do you think?

build.sbt
```diff
@@ -175,7 +175,7 @@ lazy val chill = Project(
 ).settings(sharedSettings)
   .settings(
     name := "chill",
-    crossScalaVersions += "2.13.1",
+    crossScalaVersions += "2.13.4",
```
regadas: Update this to reflect the latest `2.13.4`

README.md
regadas: Circling back on this. Does this seem ok?

nicknezis:
What is the status of this PR? Can we bump Scala from 2.13.4 to 2.13.5 and merge? Or is it blocked by something?

PR #457, opened 2020-07-15
Github actions try out

regadas:
This is the first step before introducing releases on tag push 😄

successful run in https://github.com/regadas/chill/actions/runs/170326947


regadas:
Should we give this a try? travis is becoming very unreliable these days

johnynek:
I'm +1 to migrate to github actions. I'm weakly -1 on dhall without more documentation. I don't have dhall installed, and I've never used it. I'm interested, but I'm a bit concerned that it doesn't pay it's own way to introduce to manage one yaml file which can be relatively simple and shouldn't change much.

How do you feel about removing the dhall?

regadas:
No problem at all ... I thought of removing it but then i figured i could gauge if there was interest.

PR #467, opened 2020-08-31
Backward compatible variant for migration from Scala 2.12 to 2.13

vigoo:
The default implementation in Chill does not maintain binary compatibility between Scala version 2.12 and 2.13. I guess this is an intentional decision, as even a separate test data set is provided for the two versions. 

Earlier this year we migrated a live service to Scala 2.13 that was using Chill with Akka Persistence and clustering, so had to keep binary compatibility to do a live upgrade. This PR contains an alternative `ScalaKryoInstantiator` for Scala 2.13 that preserves compatibility with 2.12 in both directions. The compatibility does not mean binary equivalence in all cases but roundtrip serialization between the two Scala versions is possible.

This was heavily tested manually and runs since months in production but there are no tests included yet. If this is something you would like to have in the core library as it may help others migrations, I'm happy to add them.

### DIFF
--- a/chill-scala/src/main/scala-2.13+/com/twitter/chill/backwardcomp/BackwardCompatibleClassResolver.scala
+++ b/chill-scala/src/main/scala-2.13+/com/twitter/chill/backwardcomp/BackwardCompatibleClassResolver.scala
@@ -1,0 +1,124 @@
+/*
+Copyright 2012 Twitter, Inc.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+
+package com.twitter.chill.backwardcomp
+
+import com.esotericsoftware.kryo.util.Util.{className, log}
+import com.esotericsoftware.kryo.util.{DefaultClassResolver, IdentityObjectIntMap, IntMap}
+import com.esotericsoftware.kryo.{Kryo, KryoException, Registration}
+import com.esotericsoftware.minlog.Log.{DEBUG, TRACE, trace}
+import com.twitter.chill.{Input, Output}
+
+class BackwardCompatibleClassResolver extends DefaultClassResolver {
+  import BackwardCompatibleClassResolver._
+
+  private val fallbackByClass: Map[Class[_], Int] =
+    Map(
+      classOf[FallbackToHashSet1] -> 23,
+      classOf[FallbackToHashSet2] -> 23,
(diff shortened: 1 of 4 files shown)


johnynek:
Thanks for the PR.

I'm willing to merge this, but I'd like to have some tests. For instance, one idea would be we could check in a file that has been serialized with one of all these items in 0.9.2 with 2.12 or something, and then assert that we deserialize it.

I get very nervous about code that people depend on that has no tests in the repo.

vigoo:
I fully agree and will add some tests!

chill-scala/src/main/scala-2.13+/com/twitter/chill/backwardcomp/BackwardCompatibleClassResolver.scala
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }
+
+  def writeFakeName[T](output: Output, name: String, placeholderType: Class[_]): Unit = {
+    output.writeVarInt(1, true)
+    if (classToNameId != null) {
+      val nameId = classToNameId.get(placeholderType, -1)
+      if (nameId != -1) {
+        output.writeVarInt(nameId, true)
+        return
```
regadas: 🎨 could we rewrite this without using `return`?

chill-scala/src/main/scala-2.13+/com/twitter/chill/backwardcomp/BackwardCompatibleExclusiveNumericRangeSerializer.scala
```diff
@@ -0,0 +1,74 @@
...
+        case "step" => field.getField.set(result, step)
+      }
+    }
+
+    result
+  }
+
+  override def write(kryo: Kryo, output: Output, `object`: T): Unit = {
+    var length: Option[Int] = None
+    for (field <- getFields) {
```
regadas: 🎨 use `while`?

chill-scala/src/main/scala-2.13+/com/twitter/chill/backwardcomp/BackwardCompatibleExclusiveNumericRangeSerializer.scala
```diff
@@ -0,0 +1,74 @@
...
+    val hashCode = input.readVarInt(false)
+    val isInclusive = input.readBoolean()
+    val last = kryo.readClassAndObject(input)
+    val num1 = kryo.readClassAndObject(input)
+    val num2 = kryo.readClassAndObject(input)
+    val numRangeElements = input.readVarInt(false)
+    val start = kryo.readClassAndObject(input)
+    val step = kryo.readClassAndObject(input)
+
+    for (field <- getFields) {
```
regadas: 🎨 use `while`? yields better perf.

regadas (commented):
Thanks for this! Would you mind rebasing as well and formatting with scalafmt? `+scalafmtAll`

vigoo:
Update: I found some issues, will push fix and tests when have time.

johnynek:
take a look at #514 --

that would break, significantly, binary compatibility, and I assume the serialized data compatibility.

you might want to weight in.

PR #473, opened 2020-09-13
README: update latest version to 0.9.5

unkarjedy:
it supports scala 2.13

PR #474, opened 2020-09-14
README.md: add note about supported scala versions

unkarjedy:
https://github.com/twitter/chill/pull/473#issuecomment-691699419

>If you have time to update the file to mention 2.13 that would be great.
Also we may have dropped 2.10, I thought we did but I'm on a phone now.


regadas (changes requested):
Thanks for this! can you remove the `2.10` reference?

regadas (approved):
Thanks

PR #556, opened 2021-05-21
Update GHA with scala 2.13.6

regadas:
After #553 this is needed. This fails due to serde compat issues with the new Vector and there's an old PR old #456 that addresses it; @johnynek If there are no objections I'll merge it and also make a new release.


regadas (commented):
Addressed by #456

Issue #529, opened 2021-03-04
java.lang.ClassNotFoundException: scala/Function0


rajeshkalaria80:
**Describe the bug**
We are using chill-akka for serialization/deserialization of remote messages. In tests, we have turned on "**akka.actor.serialize-messages**" to catch any such issues during development only. 

We have an integration tests which is failing quite often with below mentioned exception stack trace. In my troubleshooting, what I have found is that in KryoBase.scala, it fails when it is evaluating below expression and **Thread.currentThread().getContextClassLoader() is null**

```
val functions: Iterable[Class[_]] =
    (0 to 22).map { idx =>
      Class.forName("scala.Function" + idx.toString, true, Thread.currentThread().getContextClassLoader())
    }
```

[ERROR] [18:43:21.880] [t-dispatcher-20] -- **swallowing exception** during message sendakka.actor.dungeon.SerializationCheckFailedException: Failed to serialize and deserialize message of type com.test.MockParam for testing. To avoid this error, either disable 'akka.actor.serialize-messages', mark the message with 'akka.actor.NoSerializationVerificationNeeded', or configure serialization to support this message
	at akka.actor.dungeon.SerializationCheckFailedException$.apply(Dispatch.scala:27)
	at akka.actor.dungeon.Dispatch.serializeAndDeserialize(Dispatch.scala:179)
	at akka.actor.dungeon.Dispatch.sendMessage(Dispatch.scala:156)
	at akka.actor.dungeon.Dispatch.sendMessage$(Dispatch.scala:153)
	at akka.actor.ActorCell.sendMessage(ActorCell.scala:410)
	at akka.actor.Cell.sendMessage(ActorCell.scala:326)
	at akka.actor.Cell.sendMessage$(ActorCell.scala:325)
	at akka.actor.ActorCell.sendMessage(ActorCell.scala:410)
	at akka.actor.LocalActorRef.$bang(ActorRef.scala:413)
       ...
       ...
	at scala.util.Success.$anonfun$map$1(Try.scala:255)
	at scala.util.Success.map(Try.scala:213)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
[...]

PR #584, opened 2021-07-16
Fix unused functions that produce bugs (related to #529)

KitHat:
These two variables seem to be unused by the code and produce bugs like #529 
Can we omit them?

### DIFF
--- a/chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
+++ b/chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
@@ -46,30 +46,31 @@ class KryoBase(classResolver: ClassResolver, referenceResolver: ReferenceResolve
 
   protected var strategy: Option[InstantiatorStrategy] = None
 
-  val functions: Iterable[Class[_]] =
-    List(
-      classOf[Function0[_]],
-      classOf[Function1[_,_]],
-      classOf[Function2[_,_,_]],
-      classOf[Function3[_,_,_,_]],
-      classOf[Function4[_,_,_,_,_]],
-      classOf[Function5[_,_,_,_,_,_]],
-      classOf[Function6[_,_,_,_,_,_,_]],
-      classOf[Function7[_,_,_,_,_,_,_,_]],
-      classOf[Function8[_,_,_,_,_,_,_,_,_]],
-      classOf[Function9[_,_,_,_,_,_,_,_,_,_]],
-      classOf[Function10[_,_,_,_,_,_,_,_,_,_,_]],
-      classOf[Function11[_,_,_,_,_,_,_,_,_,_,_,_]],
-      classOf[Function12[_,_,_,_,_,_,_,_,_,_,_,_,_]],
-      classOf[Function13[_,_,_,_,_,_,_,_,_,_,_,_,_,_]],
-      classOf[Function14[_,_,_,_,_,_,_,_,_,_,_,_,_,_,_]],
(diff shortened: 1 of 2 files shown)


johnynek:
Thanks for the PR.

There are a few things wrong here: 1. We shouldn't use strings to get these classes. We should just make a list 23 elements long and use `List(classOf[Function0], classOf[Function1], ...`

Secondly, we shouldn't have a function like this if we aren't using it in general, BUT I think removing it will break binary compatibility (mima should fail in the CI).

I guess the easier thing it to fix the code not to use forName and reflection here, that way we fix your issue but also don't break binary compatibility.

How does that sound?

KitHat:
@johnynek sounds good, I will fix it.

johnynek:
we shouldn't need to comment out the protobuf test.

Can you merge master, and not comment that out and make sure things still pass?

chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
```diff
@@ -48,9 +47,29 @@ class KryoBase(classResolver: ClassResolver, referenceResolver: ReferenceResolve
   protected var strategy: Option[InstantiatorStrategy] = None
 
   val functions: Iterable[Class[_]] =
-    (0 to 22).map { idx =>
-      Class.forName("scala.Function" + idx.toString, true, Thread.currentThread().getContextClassLoader())
-    }
+    List(
```
johnynek: thanks for fixing this.... some of this code is pretty old and really not optimally written but also used in a variety of big systems (many versions of spark, scalding, and a few other systems) so being committed to binary compatibility really helps avoid nasty diamond dependency problems.

johnynek (changes requested):

KitHat:
Master merged, launched tests with `sbt test` -- they have passed.
Uncommented the Protobuf test as well -- it was commented out by mistake

johnynek (approved):
Thanks so much!

chill-scala/src/main/scala/com/twitter/chill/KryoBase.scala
```diff
@@ -48,9 +47,29 @@ class KryoBase(classResolver: ClassResolver, referenceResolver: ReferenceResolve
...
+      classOf[Function9[_,_,_,_,_,_,_,_,_,_]],
+      classOf[Function10[_,_,_,_,_,_,_,_,_,_,_]],
+      classOf[Function11[_,_,_,_,_,_,_,_,_,_,_,_]],
+      classOf[Function12[_,_,_,_,_,_,_,_,_,_,_,_,_]],
+      classOf[Function13[_,_,_,_,_,_,_,_,_,_,_,_,_,_]],
+      classOf[Function14[_,_,_,_,_,_,_,_,_,_,_,_,_,_,_]],
+      classOf[Function15[_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_]],
+      classOf[Function16[_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_]],
+      classOf[Function17[_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_]],
+      classOf[Function18[_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_,_]],
```
regadas: Just noticed this PR (sorry for poking); Here we jumped `Function19`

KitHat: You have an eagle eye!
Fixed this along with scalafmt

KitHat:
> Thanks so much!

I have fixed the scalafmt issue as well

PR #694, opened 2022-09-07
Update GHA actions and use Java 17

regadas:


regadas:
@johnynek do you have any thoughts regarding this?

johnynek:
sorry for the latency. This seems fine to me. Testing the past 2 LTS versions seems to make sense.

PR #799, opened 2024-04-29
akka => pekko

bharath12345:
migrating usage from akka to pekko

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,5 @@ project/build/target/
 project/plugins/target/
 project/plugins/lib_managed/
 project/plugins/src_managed/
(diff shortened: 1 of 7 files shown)


bharath12345:
@committers/@reviewers -

1. my project uses chill and we are moving from akka to pekko - hence this PR
2. UT are passing
3. am not sure if a full move to pekko for this library is desirable or you want to keep parallel releases for akka and pekko going. in case you recommend keeping parallel akka and pekko releases going then I can re-work my PR and re-submit
4. like all professional work, I have committed to some dates and releases on my project. so it will be great if I could hear back soon. Thank you!

